PR #962, opened 2015-02-17
Screwed up some basic JSP...

thongsav-usgs:


coastal-hazards-portal/src/main/webapp/index.jsp
```diff
@@ -29,7 +29,9 @@
 	String vOpenlayers = getProp("version.openlayers");
 	String vSugarJs = getProp("version.sugarjs");
 
-	String baseUrlJndiString = getProp("coastal-hazards.base.url");
+%>
```
isuftin: Why do we have two JSP blocks here?

jiwalker-usgs: One is static

isuftin: Damn code folding.

Ok.

PR #965, opened 2015-02-18
Added Marty's lil' trashcan to the bucket card

isuftin:


mwernimont:
@isuftin Looks great man

PR #973, opened 2015-02-19
Making points bigger.  Works locally, will need to tweak on dev.

jiwalker-usgs:
There is the more involved method used by other SLDs, but this might give us enough granularity.


thongsav-usgs:
jsp's as file builders.... we've been burned by this pattern in the past.

jiwalker-usgs:
We can talk about it, but it is more templating with a little bit of logic bleed in.

PR #983, opened 2015-02-24
Changed manage contents to manage bucket 

mwernimont:
changed manage contents to manage bucket in the bucket slide out dropdown on the upper right side.

Also changed the empty bucket text from "There Is Nothing In Your Bucket" to "Your bucket is empty"


isuftin:
:ok_hand:

PR #989, opened 2015-02-26
Trash can placement

mwernimont:
Used the safari developers tools to test against the iphone 4 @isuftin let me use, and found a position for the trashcan that avoids the order triangles.  

Added another order CSS rule in the min-width:992px media query to have a better looking placement of the order on larger screens


jread-usgs:
@isuftin you reviewing and merging this?

PR #992, opened 2015-02-27
First step to tree modifications

jiwalker-usgs:
Will move on to larger batches


jiwalker-usgs:
Works locally, I really want to move the JSON logic somewhere else, but take a look.

PR #997, opened 2015-03-03
Removed left alignment

isuftin:


jread-usgs:
:(

isuftin:
That shouldve said removed centering (and restored left alignment). Oopsie

PR #1001, opened 2015-03-03
Authentication conversion

thongsav-usgs:


coastal-hazards-portal/pom.xml
```diff
@@ -128,6 +139,10 @@
 		    <artifactId>jersey-mvc-jsp</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.glassfish.jersey.media</groupId>
+			<artifactId>jersey-media-moxy</artifactId>
```
jiwalker-usgs: What does moxy add?

jiwalker-usgs:
How does CIDA_AUTHORIZED_ROLE get set?

coastal-hazards-portal/pom.xml
thongsav-usgs: it's annoying, but the auth client needs it to deserialize the token coming back from the auth webservice. I believe JAXB would also work if we ever needed to switch.

thongsav-usgs:
It's arbitrarily set by the app, we will switch to using a real role, likely "CCH_ADMIN".

PR #1003, opened 2015-03-04
New media query

mwernimont:
Added a min-width media query to avoid desktop margin being picked up on phones.


mwernimont:
@isuftin Look good to you?

PR #1010, opened 2015-03-04
New animated bucket in its place

mwernimont:
I have added accessKey targets for the animations. They work if you view the svg alone on a browser, but not in the app itself. So that will require better intelligence than I have.  

The accessKeys trigger the animations as such,

addMore = s
addFirst = a
dumpBucket = w
openBucket = d
closeBucket = f


jread-usgs:
@isuftin got a min to take a peek? Those actions should be taken over by whatever you need for triggers

PR #1014, opened 2015-03-05
#1189

jiwalker-usgs:
Some of the specifics need to be worked out, but this allows for items to be templated.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
@@ -302,7 +302,7 @@ public class Item implements Serializable, Cacheable {
 	public static Item copyValues(final Item from, final Item to) {
 		Item item = new Item();
 
-		if (to.getItemType() != from.getItemType() && from.getItemType() != ItemType.template) {
+		if (to.getItemType() != from.getItemType()) {
 			throw new UnsupportedOperationException("Cannot change item type");
 		}
 		item.setId(to.getId());
@@ -319,6 +319,31 @@ public class Item implements Serializable, Cacheable {
 		item.setChildren(from.getChildren());
 		item.setDisplayedChildren(from.getDisplayedChildren());
 
+		return item;
+	}
+	
+	public Item instantiateTemplate(List<Service> withServices) {
+		Item item = new Item();
+		
+		if (this.getItemType() != ItemType.template) {
+			throw new UnsupportedOperationException("Only templates may be instantiated");
+		}
+		
+		item.setId(IdGenerator.generate());
+		item.setItemType(Item.ItemType.data);
+		
+		item.setType(getType());
+		item.setName(getName());
+		item.setAttr(getAttr());
+		item.setBbox(Bbox.copyValues(getBbox(), null));
+		item.setSummary(Summary.copyValues(getSummary(), null));
+		item.setRibbonable(isRibbonable());
+		item.setEnabled(isEnabled());
+		
+		List<Service> tmpServices = getServices();
+		tmpServices.addAll(withServices);
+		item.setServices(fillInServices(tmpServices, item.getId()));
+		
 		return item;
 	}
 
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
@@ -3,11 +3,11 @@ package gov.usgs.cida.coastalhazards.model;
 import com.google.gson.Gson;
 import gov.usgs.cida.coastalhazards.gson.GsonUtil;
 import gov.usgs.cida.coastalhazards.model.Service.ServiceType;
-import gov.usgs.cida.coastalhazards.util.ogc.WMSService;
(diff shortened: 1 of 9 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -40,6 +46,9 @@
 @PermitAll //says that all methods, unless otherwise secured, will be allowed by default
 public class ItemResource {
 
+	public static final String PUBLIC_URL = JNDISingleton.getInstance()
+			.getProperty("coastal-hazards.public.url", "http://localhost:8080/coastal-hazards-portal");
```
isuftin: I know this is a pattern we use, but it feels like there should be a defaults.properties file that we pull the default values from. Having this in the code is stanky

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
```diff
@@ -58,6 +58,7 @@
 	public enum ItemType {
 
 		aggregation,
+		template,
```
isuftin: I feel like these should be in all caps. Don't change it at this point, but isn;t that the typical enum pattern?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -0,0 +1,219 @@
...
+		catch (FileNotFoundException ex) {
+			log.error("Unable to publish shapefile", ex);
+		}
+		catch (IOException ex) {
+			log.error("Error reading stream", ex);
+		}
+		return serviceList;
+	}
+
+//	private static File shpZipToTmpFile(InputStream is) throws IOException {
```
isuftin: Please remove commented out code

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -0,0 +1,219 @@
...
+
+		File tempDirectory = FileUtils.getTempDirectory();
+		UUID uuid = UUID.randomUUID();
+		File wpsRequestFile = new File(tempDirectory, uuid.toString() + ".xml");
+		
+		try {
+
+			wpsRequestOutputStream = new FileOutputStream(wpsRequestFile);
+
+			wpsRequestOutputStream.write(new String(
```
isuftin: I bet we can get Moustache to template this out for us. I know we've been doing String concatenation in Java forever to create this XML, but this is painful

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		return layerCreated;
+	}
+
+	private static String postToWPS(String url, File wpsRequestFile) throws IOException {
+		HttpPost post;
+		HttpClient httpClient = new DefaultHttpClient();
+
+		post = new HttpPost(url);
```
isuftin: Why not instantiate on line 195? Weird

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
thongsav-usgs: I tend to agree. "reasonable defaults" end up confusing because the app works and works and works, then all of sudden doesn't when you deploy elsewhere.

jiwalker-usgs: Yeah I agree too.  We should probably move all "give me this property" calls to a common place.  Then have that use a default.properties when stuff isn't specified.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
jiwalker-usgs: Yeah, it probably should be as they are constants.  I just think they are more friendly this way.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
jiwalker-usgs: Yeah, I thought about this, didn't feel like pulling in a library at this point.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
jiwalker-usgs: Not sure, copypasta.

jiwalker-usgs:
I'm going to merge and move on with the rest of this feature.

PR #1018, opened 2015-03-05
Explore Contents Downward Arrow

mwernimont:
With the help of @isuftin the explore contents now have downward arrows in each button.  The arrows need a little CSS love but all looks good now.


mwernimont:
@jiwalker-usgs @isuftin I touched javascript so I don't wanna merge without a thumbs out.

PR #1023, opened 2015-03-09
Auth clean up, redirects working, logout button added

thongsav-usgs:


jiwalker-usgs:
Looks good, we need to get the indents figured out though.

PR #1035, opened 2015-03-13
animation and size change for bucket

mwernimont:
Changed the open and close animation to just do the subtle jiggle instead of flying away and coming back.

Also did a little CSS to make the bucket bigger and being able to click on the bucket to get in it


mwernimont:
@jread-usgs You wanna give this a look?

jread-usgs:
Looks good marty. Nice and simple.

PR #1053, opened 2015-03-23
Missed the one-to-many parameters

jiwalker-usgs:
Need to cascade the services and remove any orphans


jread-usgs:
:shipit:

jiwalker-usgs:
:rage: hibernate :rage:

PR #1097, opened 2015-04-06
removing URL from error message

mwernimont:
@isuftin Is this all I needed to do to remove the URL from the error page?


coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/server-error/index.jsp
```diff
@@ -170,7 +170,7 @@
 				// SVG wireup
 				var svg = evt.target.getSVGDocument();
 				svg.updateErrorCode('<%=request.getAttribute("javax.servlet.error.status_code")%>');
-				svg.updateErrorMessage(description, errorPath);
+				svg.updateErrorMessage(description);
```
isuftin: You will get an "undefined" error. I'd suggest 
 svg.updateErrorMessage(description, "")

mwernimont:
@isuftin That all that it needed?

PR #1104, opened 2015-04-08
Action Center Desktop Title

mwernimont:
@isuftin Would you rather do a Javascript way of doing this or you cool with CSS hiding?


isuftin:
I think this is where doing it via CSS should be perfect.

PR #1122, opened 2015-04-16
Adding cache clear mechanism

jiwalker-usgs:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+@Path(DataURI.CACHE_PATH)
+@PermitAll
+public class CacheResource {
+	
+	private static final Logger log = LoggerFactory.getLogger(CacheResource.class);
+
+	private static final String CACHE_NAME_PROP = "coastal-hazards.portal.geoserver.cache.name";
+	private static final String CACHE_NAME = "proxyCache";
+	private static final String CACHE_LOCATION_PROP = "coastal-hazards.portal.geoserver.cache.name";
```
isuftin: Why two constants with the same value? Are you looking to make them different at some point in the future?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
```diff
@@ -0,0 +1,77 @@
...
+	private static final String CACHE_LOCATION = "${java.io.tmpdir}/ehcache";
```
isuftin: How does ${} work in here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
```diff
@@ -0,0 +1,77 @@
...
+					.dynamicConfig(false)
+					.cache(ehConfig);
+			CacheManager cacheManager = CacheManager.create(managerConfig);
+			Ehcache ehcache = cacheManager.getEhcache(cacheName);
+			if (ehcache != null) {
+				ehcache.flush();
+				ehcache.removeAll();
+				response = Response.ok().build();
+			} else {
+				response = Response.serverError().entity("Error location cache to clear").build();
```
isuftin: What does this error mean?

isuftin:
This looks good to ship, but are you able to unit test this functionality? It seems like CacheResource.java L51-70 is very testable.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
jiwalker-usgs: oops, this should be cache.location

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
jiwalker-usgs: ehcache expands it

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/CacheResource.java
jiwalker-usgs: typo, meant locating

jiwalker-usgs:
I'll try to make a unit test.  This is jumping on a cache that is built elsewhere, which is the real trick here, but I should be able to make a cache and clear it.

jiwalker-usgs:
Test is fragile because the cache location needs to be the default in the cache clear case.  We should probably move to a central place where we set the cache location and use that for all usages of the cache.

PR #1126, opened 2015-04-17
Using full title rather than medium title

jiwalker-usgs:


jread-usgs:
:+1:

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/gson/adapter/ItemTreeAdapter.java
```diff
@@ -25,7 +25,7 @@ public JsonElement serialize(Item src, Type typeOfSrc, JsonSerializationContext
 			JsonObject json = (JsonObject) fullItem;
 			simplified.add("id", json.get("id"));
 			simplified.add("itemType", json.get("itemType"));
-			simplified.add("title", json.getAsJsonObject("summary").getAsJsonObject("medium").get("title"));
+			simplified.add("title", json.getAsJsonObject("summary").getAsJsonObject("full").get("title"));
```
isuftin: Ah so that's where that guy was hiding!

PR #1128, opened 2015-04-21
Trello #1276

jiwalker-usgs:
@isuftin want to take a look

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
@@ -4,10 +4,14 @@ import com.google.gson.Gson;
 import gov.usgs.cida.coastalhazards.gson.GsonUtil;
 import gov.usgs.cida.coastalhazards.jpa.ItemManager;
 import gov.usgs.cida.coastalhazards.jpa.JPAHelper;
+import gov.usgs.cida.coastalhazards.jpa.StatusManager;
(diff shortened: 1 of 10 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
```diff
@@ -118,6 +122,32 @@ public Response healthCheck() {
...
+			Status itemLastUpdate = statuses.get(StatusName.ITEM_UPDATE);
+			Status structureLastUpdate = statuses.get(StatusName.STRUCTURE_UPDATE);
+			Status cacheClearedDate = statuses.get(StatusName.CACHE_CLEAR);
+			Date itemOrStructureUpdate = null;
+			if (itemLastUpdate != null) {
+				itemOrStructureUpdate = itemLastUpdate.getLastUpdate();
+			}
+			if (structureLastUpdate != null) {
+				Date structureDate = structureLastUpdate.getLastUpdate();
+				if (structureDate != null && itemOrStructureUpdate != null && structureDate.after(itemOrStructureUpdate)) {
```
isuftin: Will an NPE be thrown here for itemOrStructureUpdate if itemLastUpdate was null?

jiwalker-usgs: I think that's what all the null checks are for.  I think I have them all covered.  Right now cacheClearedDate is the only non-null.

PR #1141, opened 2015-04-30
This should fix the saving issue, as well as a bonus dirty bit fix

jiwalker-usgs:


jiwalker-usgs:
@isuftin same as last one.

PR #1153, opened 2015-05-05
#1320

jiwalker-usgs:
Moving some download handling to service, shared between DownloadResource and ItemResource

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/service/data/DownloadService.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/service/data/DownloadService.java
@@ -16,15 +16,15 @@ import org.slf4j.LoggerFactory;
  * @author Jordan Walker <jiwalker@usgs.gov>
  */
 public class DownloadService implements AutoCloseable {
-	
+
 	private static final Logger log = LoggerFactory.getLogger(DownloadService.class);
-	
+
 	private DownloadManager manager;
-	
+
 	public DownloadService() {
 		manager = new DownloadManager();
 	}
-	
+
 	public boolean delete(String id) {
 		boolean deleted = false;
 		EntityTransaction transaction = manager.getTransaction();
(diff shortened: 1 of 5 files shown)


jiwalker-usgs:
@isuftin wish to comment?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/service/data/DownloadService.java
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}
+		return deleted;
+	}
+	
+	private boolean deleteDownload(Download download) {
+		boolean deleted = false;
+		try {
+			File stagingFolder = download.fetchZipFile().getParentFile();
+			deleted = FileUtils.deleteQuietly(stagingFolder);
```
isuftin: What happens if either this fails or manager.delete fails? I guess it's more important doe manager.delete() to pass than deleting @ the file system. Maybe flip this so manager.delete() runs first?

jiwalker-usgs: Yeah, that is probably smart.  I'm a bit conflicted about how to fail well here, since we do want all rows removed and files deleted.

isuftin: I think it's better to just orphan the files but make sure we remove them from the database rather than the other way around. Flipping the two will accomplish this, yeah?

isuftin: We could probably also create a shutdown-time sweeper to check the file system and cross-check the db and remove directories and files not in the db

jiwalker-usgs: Yeah, I was thinking the same thing.

PR #1161, opened 2015-05-07
Switched to working with LESS. Will be converting more code soon

isuftin:


jiwalker-usgs:
How are we compiling the less?

isuftin:
I have Netbeans compiling it on save

isuftin:
I should probably get a maven less compiler in there as well

PR #1177, opened 2015-05-08
Info.less variables added

mwernimont:


mwernimont:
@isuftin yes?

coastal-hazards-portal/src/main/webapp/css/info/info.css
```diff
@@ -70,7 +70,7 @@ ul {
   margin: 15px 0px 5px 10px;
   padding-left: 15px;
   font-size: 1.2em;
-  color: #303030;
+  color: #333333;
```
isuftin: Note this change. This is probably not what you want

coastal-hazards-portal/src/main/webapp/css/info/info.css
```diff
@@ -176,15 +176,14 @@ h1.mobile {
     padding-top: 5px;
   }
   .icon h3 a {
-    color: #0f416f;
+    color: #003366;
   }
   .icon h3 p {
-    color: #3b3b3b;
+    color: #333333;
```
isuftin: Also note this change.

coastal-hazards-portal/src/main/webapp/css/info/info.css
```diff
@@ -176,15 +176,14 @@ h1.mobile {
     padding-top: 5px;
   }
   .icon h3 a {
-    color: #0f416f;
+    color: #003366;
```
isuftin: Are you sure?

coastal-hazards-portal/src/main/webapp/less/info/info.less
```diff
@@ -6,9 +6,11 @@
 	border:1px solid @cch-grey;
 }
 
+@infoBorder:1px solid rgb(219,219,219);
+
 .footerbar-link() {
 	margin-right: 40px;
-	color: #ffffff;
+	color: @white;
```
isuftin: I'd suggest that we probably don't need this variale since CSS has built-in variables for basic colors.

So... 

color: white

...also works

coastal-hazards-portal/src/main/webapp/css/info/info.css
mwernimont: I am sure about that, it was a blue that was only a few shades away from our #003366, I don't remember my reasons for doing that and they probably weren't good ones in the first place

isuftin: Ok. I didn't want this to be due to only having 3 digits of a hex code instead of all 6 screwing you over

isuftin:
Aside from the comments I made, you should be good to go. Merge at will

coastal-hazards-portal/src/main/webapp/css/info/info.css
mwernimont: @isuftin I'm just taking the rando gray color I picked for this page and making them match our text colors throughout the app.  I for some reason was doing really weird stuff back in the day.  I made this when I first started right?

isuftin: Yupyup

coastal-hazards-portal/src/main/webapp/less/info/info.less
mwernimont: True, but thought we wanted the ability to change all aspects in one switch over going to each element and changing it...only reason I did it

PR #1193, opened 2015-05-11
fix for sysdata export

jread-usgs:


jread-usgs:
@jiwalker-usgs review? didn't mean to merge...but will keep unless you have issue w/ it.

coastal-hazards-n52/src/main/webapp/R/r_scripts/item.summary.R
```diff
@@ -4,9 +4,9 @@
 
 # find theme
 
-library(itemSummaryService)
+library(hazardItems)
 library(RCurl)
-
+attrMap <- hazardItems:::attrMap
```
jiwalker-usgs: https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-dblcolon.html says this is generally not a good idea.  This is probably fine since we own the package as well, but something to think about.

PR #1194, opened 2015-05-11
Spinners fixed for real this time

isuftin:


mhines-usgs:
for reals?

isuftin:
I think so :/

PR #1195, opened 2015-05-12
LESS rework

isuftin:


coastal-hazards-portal/src/main/webapp/less/cch.less
```diff
@@ -1,4 +1,7 @@
-@import "webjars/META-INF/resources/webjars/bootstrap/3.3.4/less/bootstrap.less";
+#bootstrap() {
```
mbucknell: Will you now need to refer to bootstrap mixins like this: #bootstrap > .btn? The advantage to namespacing is you know where the mixin comes from.

isuftin: This serves two main advantages here. One is the namespacing advantage that you've described. The other is that doing the import in this way does not cause the Netbeans compiler to barf Bootstrap CSS into every CSS file I import cch.less into. So the second advantage is also a workaround :(

PR #1200, opened 2015-05-12
Found bug poking around on dev

jiwalker-usgs:


isuftin:
How...was this working?

PR #1207, opened 2015-05-14
Expose attrs

jiwalker-usgs:


jiwalker-usgs:
@isuftin I put in a function for you to get the storm attributes.  Let me know if you end up needing that ordered.

PR #1228, opened 2015-05-18
First attempt to fix display of template item

isuftin:


jiwalker-usgs:
Looks good to me :shipit:

PR #1269, opened 2015-06-04
HACK ALERT: Changing item type of storm track child item's parent to …

isuftin:
…storm_track in a very hackish way.


jiwalker-usgs:
Might have a problem here because TRACK isn't a normal type and it might screw stuff up elsewhere, but it seems ok.

PR #1281, opened 2015-06-09
Autofocus on username field on login page

mhines-usgs:
because no more mousin'


mhines-usgs:
my life just got 100% better because of autofocus.

isuftin:
Remember that it's only supported in IE10 and up in that IE family :D

mhines-usgs:
It supports me.

PR #1282, opened 2015-06-09
Fix internet explorer breaking due to SVG animations

isuftin:


coastal-hazards-portal/src/main/webapp/images/banner/bucket/animated_bucket.svg
```diff
@@ -42,24 +49,24 @@
...
 							function bucketOpen(){
 								var ele = document.getElementById('jiggle1');
-								ele.beginElement();
+								beginElement(ele);
 							}
 								
 							function bucketClose(){
 								var ele = document.getElementById('jiggle1');
-								ele.beginElement();
+								beginElement(ele);
```
jread-usgs: so this is the preferred `begin()`?

PR #1286, opened 2015-06-09
#1350

jiwalker-usgs:
quick win

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DownloadResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DownloadResource.java
@@ -208,6 +208,10 @@ public class DownloadResource {
 	public Response deleteStagedItem(@PathParam("id") String itemId, @Context HttpServletRequest request) {
 		Response response = null;
 		try (DownloadService downloadService = new DownloadService()) {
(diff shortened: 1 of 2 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DownloadResource.java
```diff
@@ -208,6 +208,10 @@ public Response displayStagedItems() {
 	public Response deleteStagedItem(@PathParam("id") String itemId, @Context HttpServletRequest request) {
 		Response response = null;
 		try (DownloadService downloadService = new DownloadService()) {
+			Download download = downloadService.get(itemId);
+			if (download == null) {
```
isuftin: Is there any reason to throw an exception from here? You could just give the same response as if it were deleted, no?

jiwalker-usgs: It really isn't deleted since it was never there.  We should report that they are trying to delete something that isn't there.

isuftin: That's fair. I figured since the outcome was the same, why complicate the process? But this works as well.

PR #1309, opened 2015-06-17
JS cleaning and 304 fix

isuftin:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -73,10 +73,28 @@ public Response getItem(@PathParam("id") String id,
...
+					
+					modified = lastItemUpdateDate.after(lastStructureUpdateDate) ? lastItemUpdateDate : lastStructureUpdateDate;
+				} else {
+					// At least one of the two do not exist, so find out if at 
+					// least one exists and use that. 
+					if (lastItemUpdate != null) {
+						modified = lastItemUpdate.getLastUpdate();
+					}
+					if (lastStructureUpdate != null) {
+						modified = lastStructureUpdate.getLastUpdate();
```
jiwalker-usgs: I think we need to figure out which is newer and set that to modified (currently structure will always win).

jiwalker-usgs: Nevermind, this is only if one of them is null.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -157,7 +165,7 @@ private boolean retainAggregations(JsonObject parent) {
 		while (iterator.hasNext()) {
 			
 			String attr = "";
-			Layer layer = null;
```
jiwalker-usgs: I like setting null to be explicit.  I know it will always be set, but that is just my preference.

isuftin: Want me to re-add that? ;)

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -257,8 +265,9 @@ private Summary makeSummary(Layer layer, String attr) throws JsonSyntaxException
 	}
 	
 	private Map<String, Item> makeChildItemMap(List<Item> children) {
-		Map<String, Item> result = new LinkedHashMap<>();
+		Map<String, Item> result = new LinkedHashMap<>(0);
```
jiwalker-usgs: Why are you doing this?  Is it because the hash needs an approximate size?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
isuftin: <u style="color:yellow">Layer layer = null</u>

:D

jiwalker-usgs: It's fine, just felt like I should justify myself.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
isuftin: Yeah. Quick optimization win even though it matters not at all

jiwalker-usgs:
Looks solid :+1:

PR #1320, opened 2015-06-24
Staged downloads fix (1 of ?)

isuftin:


jiwalker-usgs:
This looks good, I think I had some weird logic for not existing, but that might have broke with some refactor.  I don't think this will actually fix it, since the problem is on realtime which would not exist as staged downloads in database.

PR #1328, opened 2015-06-26
If staging is not sucessful for an item, update the database to reflect this and pass a 501 to the front

isuftin:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DownloadResource.java
```diff
@@ -71,7 +72,9 @@ public Response checkItemAvailability(@PathParam("id") String id) throws IOExcep
 
 					// Check if the file location in the database 
 					// If it is null or blank, the download has been accepted
-					if (StringUtils.isBlank(persistenceURI)) {
+					if (download.isProblem()) {
+						response = Response.status(NOT_IMPLEMENTED).build();
```
jiwalker-usgs: I'm not sure if isProblem necessarily indicates NOT_IMPLEMENTED.  I think there are cases where there is a problem in building the download outside of it not being implemented.

jiwalker-usgs:
I think the not_implemented should be triggered before it touches the database or tries to build the download.  Otherwise if there is an issue we should return 500.

isuftin:
How do we decide if there's a problem without ever touching the database?

jiwalker-usgs:
It really only applies to single data item downloads.  If there is no source_wfs, downloads are not supported.

isuftin:
At this point I'm not sure what you are envisioning the implementation to be on the back-end. Should talk in person.

PR #1334, opened 2015-06-30
Added much logging for checking/staging downloads

isuftin:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DownloadResource.java
```diff
@@ -52,7 +52,7 @@
 @PermitAll
 public class DownloadResource {
 
-	private static final Logger log = LoggerFactory.getLogger(DownloadResource.class);
```
jiwalker-usgs: I'm a fan of lower case log, seems more friendly to me.

isuftin: I couldn't tell which would be best. I looked at a few other files in the vicinity with upper case LOG so I rolled with it ;)

PR #1345, opened 2015-07-14
Refactoring head/get download requests and staging

isuftin:
Trying to isolate try-with-resources blocks only to where they're needed in an attempt to shorten open db connections


isuftin:
I will put this on a branch. I have another PR that's a lot more pressing coming in

PR #1347, opened 2015-07-14
Trying to isolate try-with-resources blocks only to where they're needed in an attempt to shorten open db connections

isuftin:


coastal-hazards-export/src/main/java/gov/usgs/cida/coastalhazards/export/HttpComponentsWFSClient.java
```diff
@@ -51,7 +51,7 @@ public void setupDatastoreFromEndpoint(String wfsUrl) throws MalformedURLExcepti
 
     @Override
     public SimpleFeatureCollection getFeatureCollection(String typeName, Filter filter) throws IOException {
-        SimpleFeatureCollection collection= null;
+        SimpleFeatureCollection collection;
```
jiwalker-usgs: Stop doing this to my beautiful code.  Don't let netbeans do the coding for you :(

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/download/DownloadUtility.java
```diff
@@ -73,12 +73,26 @@ public synchronized static void unlock(File file) {
 		}
 	}
 
+	/**
+	 * Find where there staging directory is located for all download items.
```
jiwalker-usgs: Thanks for documenting this stuff, I should probably be better about that.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/download/DownloadUtility.java
```diff
@@ -324,36 +339,49 @@ private static WFSService getWfsService(Item item) {
 
 		@Override
 		public Download call() throws IOException {
```
jiwalker-usgs: I like the idea of smaller try-blocks, but in practice I think it is making the code harder to read.

PR #1363, opened 2015-07-21
Create intro for back of card

isuftin:


mhines-usgs:
Love it. :shipit: :+1:

PR #1364, opened 2015-07-21
Ensuring no session writing during tour

isuftin:


mhines-usgs:
Works awesome. Thanks for clearin' up that bucket.

PR #1386, opened 2015-07-23
favico and connection plugging

isuftin:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
```diff
@@ -30,7 +32,7 @@
 @Path(DataURI.HEALTH_PATH)
 @PermitAll //says that all methods, unless otherwise secured, will be allowed by default
 public class HealthResource {
-	
+	private static final Logger LOG = LoggerFactory.getLogger(HealthResource.class);
```
jiwalker-usgs: Just going to note my objection to LOG vs log.

PR #1426, opened 2015-10-10
Wps refactor

isuftin:


isuftin:
Would be good to get your thoughts on this. I only left the ribboning process and the PlaintextPPIO. It doesn't seem we're using the PPIO in this project but because there's no hard link, I couldn't be sure.

PR #1427, opened 2015-10-10
Commons refactor

isuftin:


isuftin:
@jiwalker-usgs Would be good for you to look through this. This should be merged *after* https://github.com/USGS-CIDA/coastal-hazards/pull/1426

PR #1432, opened 2015-10-16
Removing the USGS svg overlay from map

mhines-usgs:
because we now have their logo in the banner


mhines-usgs:
https://trello.com/c/pfWU2mhF/50-1-as-john-haine-i-want-to-replace-the-cida-cmgp-logo-with-a-usgs-logo-so-that-we-re-cooler-3

PR #1441, opened 2016-06-06
Adding lower bound to dunes, extreme, and mean water levels

jiwalker-usgs:
* SLD jsp is now aware of bounded or unbounded lower edge


mhines-usgs:
:+1: let's test on dev.

PR #1443, opened 2016-06-13
icons changed

mwernimont:
Made the circle smaller and the the blue line thinner. My local version wouldn't uncache to show me the changes of the icons, so please let me know if they don't change for you as well. @mhines-usgs


mhines-usgs:
perfect!!

mhines-usgs:
fyi, @mwernimont you can peep these on github, if you go to Files Changed, then click on "View" on the right side, it will actually render the graphic instead of the raw svg.

mwernimont:
I call hacks! lol But nice, was unaware of thee ability. Makes my job easier. :)

PR #1444, opened 2016-06-22
Docker updates that never got merged

isuftin:


mhines-usgs:
@isuftin should these changes be merged, then?

PR #1445, opened 2016-08-03
Share Bucket to Share View

mwernimont:
@mhines-usgs Chose the Share View option, the share bucket view would get a little long for mobile and I always choose less words to make designs cleaner.


mhines-usgs:
shanks marty, I'm going to merge and check it out on dev.

PR #1451, opened 2016-09-15
ae static legend image

mwernimont:
@jiwalker-usgs Look like what you thought? 

Took a screen shot of an existing legend pice and used it for size and color shape. Can always edit it if its not quite right.


jiwalker-usgs:
This doesn't look right to me, I think we want a ramp rather than bins.  I'll take another look at the ticket though.

mhines-usgs:
AE has bins, there are three different ones... you're probably thinking of the probability of AE which is a ramp...

mhines-usgs:
http://woodshole.er.usgs.gov/project-pages/coastal_response/images/AE_full.jpg
http://woodshole.er.usgs.gov/project-pages/coastal_response/images/PAE_full.jpg
http://woodshole.er.usgs.gov/project-pages/coastal_response/images/CR_full.jpg

jiwalker-usgs:
if it has bins, we should use the current mechanism, what we don't have now is a way to do a ramp.

mhines-usgs:
Agh, so bins == java/jsp way, what Marty should be doing is the ramp? If so, that's my fault for confusing you Marty...

mwernimont:
I have no idea what ramp or bin is, so I'm at a loss of what has been discussed here

PR #1467, opened 2016-10-03
Fixing GeoServer's server.xml

cschroed-usgs:


cschroed-usgs:
After merging this, the xml parsing error goes away. Now running into:
```
java.io.FileNotFoundException: /usr/local/tomcat/ssl/.keystore (No such file or directory)
```

PR #1470, opened 2016-10-04
Parameterizing tomcat files

cschroed-usgs:


.gitignore
```diff
@@ -41,3 +41,4 @@ dependency-reduced-pom.xml
 
 **/.DS_Store
 
+Docker/compose_local.env
```
isuftin: We don't necessarily mandate `_local` and are leaving it free to be anything. Wonder if we can do Docker/compose[:: any alphanumeric::].env for the ignore

cschroed-usgs: good call!

PR #1471, opened 2016-10-04
Ignoring a broader range of compose*.env files.

cschroed-usgs:


cschroed-usgs:
I verified this:
```
$> cp compose.env compose_wombat.env
$> git add compose_wombat.env 
The following paths are ignored by one of your .gitignore files:
Docker/compose_wombat.env
Use -f if you really want to add them.
```

PR #1475, opened 2016-10-14
Improving postgres container parameterization

cschroed-usgs:
I figured out that you can github codeload tarred branches as well as tags. I stopped hardcoding a tag prefix. I switched the default liquibase scripts to run to master. I created an example in case someone wanted to rely on a tag.


cschroed-usgs:
Don't merge this yet -- it's not clear that parameterization is working. Further local testing shows that it is disregarding the user-specified value, and instead using the default value.

cschroed-usgs:
Closing -- the parameterization pattern we are using isn't working for other images, so we should not spread it to another image.

PR #1483, opened 2016-10-19
Cchs 225 pull raster from elsewhere wps process

cschroed-usgs:
This is just the WPS component. Other work will follow on to further integrate this process into the system.


coastal-hazards-wps/pom.xml
```diff
@@ -13,6 +13,10 @@
 	<name>Coastal Hazards WPS Processes</name>
 
 	<dependencies>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-all</artifactId>
+            </dependency>
```
cschroed-usgs: The version is inherited from GeoServer.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/UnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+    static final Logger LOGGER = Logging.getLogger(UnzipProcess.class);
+    static final String TOKEN_PROPERTY_NAME = "gov.usgs.cida.coastalhazards.wps.unzip.process.token";
+    static final String UNZIP_BASE_PROPERTY_NAME = "gov.usgs.cida.coastalhazards.wps.unzip.process.unzip.base";
+    private DynamicReadOnlyProperties properties;
+    private HttpClient httpClient;
+    
+    @DescribeResult(name = "filePath", description = "path to the directory where the file was unzipped")
+    public List<String> execute(
+            @DescribeParameter(name = "zipUrl", min = 1, max = 1, description = "URL to the zipped file to retrieve") String zipUrl,
+            @DescribeParameter(name = "token", min = 1, max = 1, description = "Token for authorizing the upload") String token
```
cschroed-usgs: As far as I can tell, GeoServer doesn't offer WPS per-process security until version 2.7.4. Since we are on 2.4, we must roll our own.

cschroed-usgs: The token will need to be added to the context.xml for both the GeoServer and the Portal tomcat instances.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/UnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+            throw new ProcessException("error getting next entry in zip file" , ex);
+        } finally {
+            IOUtils.closeQuietly(zipStream);
+        }
+        return unzippedFiles;
+    }
+    /**
+     * @return the properties, defaulting to JNDI
+     */
+    DynamicReadOnlyProperties getProperties() {
```
cschroed-usgs: Using JNDISingleton as a last resort in this accessor made the class easier to test this class -- I got a bunch of errors when I tried to call the method in a test environment without an application.json.

coastal-hazards-wps/src/test/java/gov/usgs/cida/coastalhazards/wps/UnzipProcessTest.java
```diff
@@ -0,0 +1,247 @@
...
+        instance.getNewZipDestination(null);
+    }
+    
+    @Test
+    public void testGetNewZipDestination(){
+        String zipDestination = instance.getNewZipDestination("asdf");
+        assertTrue("path string should be non-empty", 0 < zipDestination.length());
+    }
+    
+    public HttpClient mockHttpClient(int code, InputStream content){
```
cschroed-usgs: Mockito was a big win here -- went from ~300 lines to ~30.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+    static final Logger LOGGER = Logging.getLogger(FetchAndUnzipProcess.class);
+    static final String TOKEN_PROPERTY_NAME = "gov.usgs.cida.coastalhazards.wps.unzip.process.token";
+    static final String UNZIP_BASE_PROPERTY_NAME = "gov.usgs.cida.coastalhazards.wps.unzip.process.unzip.base";
+    private DynamicReadOnlyProperties properties;
+    private HttpClient httpClient;
+    
+    @DescribeResult(name = "filePath", description = "path to the directory where the file was unzipped")
+    public List<String> execute(
+            @DescribeParameter(name = "zipUrl", min = 1, max = 1, description = "URL to the zipped file to retrieve") String zipUrl,
+            @DescribeParameter(name = "token", min = 1, max = 1, description = "Token for authorizing the upload") String token
```
cschroed-usgs: As far as I can tell, GeoServer doesn't offer WPS per-process security until version 2.7.4. Since we are on 2.4, we must roll our own. The token will need to be added to the context.xml for both the GeoServer and the Portal tomcat instances.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+        
+        if(isAuthorized(token)){
+            ZipInputStream zipStream = getZipFromUrl(zipUrl, getHttpClient());
+            unzippedPaths = unzipToDir(zipStream, getNewZipDestination());
+        } else {
+            throw new ProcessException( new SecurityException("Not Authorized."));
+        }
+        return unzippedPaths;
+    }
+    boolean isAuthorized(String token){
```
jiwalker-usgs: I'm a bit on the fence about the default (package protected) visibility here.  I get that it is for the unit tests, but these just feel like private methods.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+            expectedToken = null;
+            authorized = true;
+        }
+        return authorized;
+    }
+    /**
+     * Everything except for tests should call this method to get a new zip destination
+     * @return a unique path
+     */
+    String getNewZipDestination(){
```
jiwalker-usgs: Did you consider returning a `File` rather than a `String` here?  It looks like you are only using the result in the context of a file, so it might be cleaner to make the `File` at this point.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+    
+    List<String> unzipToDir(ZipInputStream zipStream, String zipDir) {
+        List<String> unzippedFiles = new ArrayList<>();
+        try {
+            ZipEntry entry;
+            while (null != (entry = zipStream.getNextEntry())) {
+                if(!entry.isDirectory()){
+                    String entryFileName = entry.getName();
+
+                    File entryFile = new File(zipDir + File.separator + entryFileName);
```
jiwalker-usgs: If you get zipDir as a `File`, this can be changed to `new File(zipDir, entryFileName)` and it handles the File.separator deal for you.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    void setProperties(DynamicReadOnlyProperties properties) {
+        this.properties = properties;
+    }
+
+    /**
+     * @return the httpClient
+     */
+    HttpClient getHttpClient() {
+        return null == httpClient ? new DefaultHttpClient() : httpClient;
```
jiwalker-usgs: Just style, but I like wrapping conditionals in `()` for this statement.  In my view it improves readability in the ternary operator that is already somewhat tricky.

jiwalker-usgs: Also, I haven't looked into it, but does the http client need to have a lifecycle to deal with connection pools and the sort?  I know this class only gets instantiated once, so this could possibly get stale.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
cschroed-usgs: I didn't. I will do that. Thanks!

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
cschroed-usgs: Nice!

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
cschroed-usgs: I like the style tip. I will do that.

cschroed-usgs: I'm not sure about connection pooling etc. I only expect someone to run the package-private setHttpClient method in tests. If no one is using that method in production, then the result of getHttpClient() is always a freshly-instantiated client built for each WPS Execute request.

cschroed-usgs:
@jiwalker-usgs I closed this by accident when I renamed some branches. I re-opened a pull request for the new branch name: https://github.com/USGS-CIDA/coastal-hazards/pull/1487, but we can continue the review right here. Does this need any additional work before it is mergeworthy?

jiwalker-usgs:
I think it is mergeworthy, the only thing remaining is the visibility of the functions.  One option is to make them `protected` and have the test class "extend" this class in order to gain access.  I don't think that solution is much better than using the default, so wouldn't push it.  This was one thing I really liked about the spock testing framework was that it did the reflection bit for you (groovy has some fanciness around this), but it was more of a pain than it was worth.

cschroed-usgs:
Thanks Jordan. It's been a while since I've tried the approach you describe. Since neither approach is  stupendous, and the package-private way is already implemented, I'm inclined to merge as-is. I would like to try the approach you advocated next time I come across a similar situation so that I can compare+contrast.

PR #1484, opened 2016-10-20
updates from code review comments

smlarson-usgs:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
```diff
@@ -11,20 +11,14 @@
...
-	protected static final String units = "Inundate to Dynamic"; 
+public final class RasterCR {
+    
+        private static final String[] attrs = {CR};
+        protected static final String jspPath = "/SLD/raster_cr.jsp";
+        protected static final String units = "percent"; 
         protected static final List<Map<String, Object>> bins;
-        
-        // type = intervals for the ColorMap
         protected static final float[] thresholds = {0.0f, .31f, .66f, .95f}; //used to evaluate what color the pixel should receive. (index)
```
jiwalker-usgs: These were just the values from the data range, to be more general these should be .33, .67, and 1.0

PR #1487, opened 2016-10-24
Cchs 225 wps fetch and unzip

cschroed-usgs:


cschroed-usgs:
This is the same as https://github.com/USGS-CIDA/coastal-hazards/pull/1483. Created this because I had to rename branches.

jiwalker-usgs:
made comment on other review, this is good to merge.

PR #1488, opened 2016-10-28
Cchs 223

cschroed-usgs:
I created the concept of a 'domain' in SLD Info. A domain describes whether the visualization will be discrete or continuous. I then vary the legend rendering behavior according to the 'domain' attribute that the server specifies. On the client I created a generic function for rendering continuous gradient legends. The generic function supports caller-driven customization of bin labels. I implemented a customization of the bin labeling for the CR data set.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Domain.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Domain.java
@@ -2,6 +2,8 @@
 package gov.usgs.cida.coastalhazards.sld;
 
 enum Domain {
-    DISCRETE,
-    CONTINUOUS
+	//These exact values are hardcoded in the JS of the web client.
+	//If you change them here, you must update them in the web client.
+	DISCRETE,
+	CONTINUOUS
 }
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Domain.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Domain.java
@@ -1,0 +1,7 @@
+
+package gov.usgs.cida.coastalhazards.sld;
+
+enum Domain {
+    DISCRETE,
+    CONTINUOUS
+}
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
@@ -36,6 +36,6 @@ public final class RasterCR {
 	}
 
 	public static final SLDConfig rasterConfig = new SLDConfig(
-			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, Domain.CONTINUOUS
+			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, DataDistribution.CONTINUOUS
 	);
 }
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
@@ -36,6 +36,6 @@ public final class RasterCR {
 	}
 
 	public static final SLDConfig rasterConfig = new SLDConfig(
-			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, DataDistribution.CONTINUOUS
+			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, LegendType.CONTINUOUS
 	);
 }
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
@@ -36,6 +36,6 @@ public final class RasterCR {
 	}
 
 	public static final SLDConfig rasterConfig = new SLDConfig(
-			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, attrs, thresholds, colors, bins
-	);    
+			jspPath, units, SLDGenerator.style, SLDGenerator.STROKE_WIDTH_DEFAULT, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, Domain.CONTINUOUS
+	);
 }
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
@@ -15,7 +15,7 @@ public final class RasterCR {
     
         private static final String[] attrs = {CR};
         protected static final String jspPath = "/SLD/raster_cr.jsp";
-        protected static final String units = "percent"; 
+        protected static final String units = "likelihood"; 
         protected static final List<Map<String, Object>> bins;
         protected static final float[] thresholds = {0.0f, .33f, .67f, 1.0f}; //used to evaluate what color the pixel should receive. (index) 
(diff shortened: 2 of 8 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Domain.java
```diff
@@ -0,0 +1,7 @@
+
+package gov.usgs.cida.coastalhazards.sld;
+
+enum Domain {
```
cschroed-usgs: Should this enum define client-side values for the Domain too, or is it fine for there to be a hard coupling across server and client for these values?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDConfig.java
```diff
@@ -56,10 +56,12 @@
      * @param bins 
      */
 	public SLDConfig(String jspPath, String units, String style, int strokeWidth, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins) {
-        this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins);
+            this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, SLDGenerator.DOMAIN_DEFAULT);
 	}
-    
     public SLDConfig(String jspPath, String units, String style, int strokeWidth, float strokeOpacity, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins) {
```
cschroed-usgs: I made a commonly-used constructor default to discrete. This default can be overriden via the most verbose constructor.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -78,8 +78,70 @@ CCH.Objects.Widget.Legend = function (args) {
...
+		var numberOfNonTerminalBins = bins.length - 2;
+		var gradientIncrement = 100 / (1 + numberOfNonTerminalBins);
+		var binsForTemplate = bins.sort(sortByLowerBound).map(function(bin, index, bins){
+			var percent = index * gradientIncrement;
+			return {
+				'color': bin.color,
+				'percent': percent,
+				'binLabel': binLabeler(bin, index, bins)
+			};
+		});
```
cschroed-usgs: @mwernimont this is where the template is called from the JS.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -78,8 +78,70 @@ CCH.Objects.Widget.Legend = function (args) {
...
+		var $legendElt = $(CCH.Objects.Widget.Legend.prototype.templates.continuous({
+			browserSpecificGradients : [
```
cschroed-usgs: @mwernimont thanks for researching these CSS tricks!

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -337,17 +401,70 @@ CCH.Objects.Widget.Legend = function (args) {
...
+	 * @param {Object} sld
+	 * @param {Object} item
+	 * @returns {Object} an object to pass to a legend renderer function
+	 */
+	me.customizeLegendRendererArguments = function(sld, item) {
+		var legendRendererArguments = {
+			sld: sld,
+			item: item
+		};
+		if("CR" === item.attr){
```
cschroed-usgs: Is this an appropriate condition for customizing? Is it too general? Too specific?

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -337,17 +401,70 @@ CCH.Objects.Widget.Legend = function (args) {
...
 		var sld = args.sld,
 			item = args.item,
 			index = args.index,
 			attr = item.attr,
-			$legendTable = me.generateGenericLegendTable({
-				sld: sld,
-				item : item
-			});
+                        $legendTable;
+			var domainToLegendRenderer = {
```
cschroed-usgs: This is a client-side example of the usage of the Domain.java enum values.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -641,16 +758,28 @@ CCH.Objects.Widget.Legend = function (args) {
 
 (function () {
 	"use strict";
```
cschroed-usgs: I reworked this so that it could easily load many templates instead of just one.

coastal-hazards-portal/src/main/webapp/resource/template/handlebars/legend/continuous.html
```diff
@@ -0,0 +1,22 @@
+<table class="continuous_legend">
+  <tr>
+    <td class="continuous_legend_color_ramp" style="
+	background: red; /* For browsers that do not support gradients */
```
cschroed-usgs: @mwernimont : this template loopage makes it so we can concisely support all the browser-dependent fixes, for arbitrary numbers of bins.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
jiwalker-usgs: I think in theory this would come from the JSON "sldinfo" call and this text should be built in the SLD descriptor.  It is possible that the structure for domain=continuous is different than discrete, and you can pull the text in that manner.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
jiwalker-usgs: this would be where this could somehow be jsp'd in, seems complicated though.

coastal-hazards-portal/src/main/webapp/resource/template/handlebars/legend/continuous.html
jiwalker-usgs: fancy :dancer:

cschroed-usgs: :smile: thanks!

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
cschroed-usgs: > I think in theory this would come from the JSON "sldinfo" call and this text should be built in the SLD descriptor. 
Bingo! Yep, that's what happens.
> It is possible that the structure for domain=continuous is different than discrete, and you can pull the text in that manner.
I thought about conditioning on domain, but I think it is quite possible for a domain=continuous sldInfo to need  labels that aren't specifically ('100% Dynamic', '&nbsp;' and '100% Inundate'). It seems like the labeler I'm building below depends more on the real-world meaning of this specific raster, and less on whether the values in the raster are restricted to discrete values, or spread over a continuous range.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDConfig.java
```diff
@@ -56,10 +56,12 @@
      * @param bins 
      */
 	public SLDConfig(String jspPath, String units, String style, int strokeWidth, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins) {
-        this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins);
+            this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, SLDGenerator.DOMAIN_DEFAULT);
```
smlarson-usgs: The number of args is getting rather large for the constructor. Is there a container object or two that could be passed instead of repeatedly widening the signature?

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -8,6 +8,13 @@
 window.CCH = CCH || {};
 CCH.Objects = CCH.Objects || {};
 CCH.Objects.Widget = CCH.Objects.Widget || {};
+
```
cschroed-usgs: @jiwalker-usgs - The JSP route was pushing back, so I went with this halfway approach instead. The map creates a single source of truth for client-side legend types. If future devs need to change that hard client-server coupling, they only have to do it in this map, as opposed to strings littered throughout the code.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterCR.java
```diff
@@ -15,7 +15,7 @@
     
         private static final String[] attrs = {CR};
         protected static final String jspPath = "/SLD/raster_cr.jsp";
-        protected static final String units = "percent"; 
+        protected static final String units = "likelihood";
```
cschroed-usgs: I think it's both. If not, it would be a quick change to do after user testing.

PR #1491, opened 2016-11-01
Fix for GeoServer GRAY_INDEX attribute bug

cschroed-usgs:
This hack presumes that a layer is a raster layer if it detects that the only attribute on a feature is  GRAY_INDEX. This is dependent on a bug in GeoServer 2.4, or on us continuing to name the attribute in our raster layers GRAY_INDEX when we upgrade to a newer version of GeoServer.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
@@ -4,7 +4,32 @@
 /*global LOG*/
 /*global CCH*/
 /*global OpenLayers*/
-CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, {
+CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, 
+(function(){
+	/**
+	 * If the features are from a vector data source, return 'attrName', 
+	 * else if the features are from a raster source, override them with the
+	 * only attribute name that GeoServer 2.4 supports for raster layers.
+	 * 
+	 * @param {Array<Object>} features
+	 * @param {String} attrName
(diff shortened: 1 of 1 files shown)


mhines-usgs:
FYI @cschroed-usgs  The raster data coming in from the cooperators is not actually named GRAY_INDEX, but rather Geoserver presumes/overrides it on import or something.. Not sure if that matters or not. It is called something else in their original files.

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -4,7 +4,32 @@
 /*global LOG*/
 /*global CCH*/
 /*global OpenLayers*/
-CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, {
+CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, 
+(function(){
```
cschroed-usgs: Switched this from an object literal to an IIFE so that I could encapsulate the overrideAttributeName function via the Module Pattern.
http://benalman.com/news/2010/11/immediately-invoked-function-expression/

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -382,7 +407,7 @@ CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFea
 				});
 			},
 			isMissing = function(val) {
-				return isNaN(val) || val === -999;
+				return isNaN(val) || val === -999 || val === -3.4028234663852886e+38;
```
cschroed-usgs: The new raster has a new 'missing' value.

cschroed-usgs:
@mhines-usgs Yep! This fix works relies on that.

PR #1494, opened 2016-11-02
Trying a fix closer to the status quo

cschroed-usgs:
This may eliminate regressions encountered when clicking on non-AE layers.


cschroed-usgs:
Going to merge so that I can test this with the wide variety of data sources on dev. I don't want to set them up locally right meow.

PR #1495, opened 2016-11-02
Cchs 243

cschroed-usgs:
Creating a really skeletal upload page for both metadata and zipped tiff files.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
@@ -72,6 +72,14 @@ public class PublishResource {
     @RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
     @GET
     @Produces(MediaType.TEXT_HTML)
+    @Path("/item/raster")
+    public Response createRasterItem(@Context HttpServletRequest req) throws URISyntaxException {
+       return Response.ok(new Viewable("/WEB-INF/jsp/publish/item/raster.jsp", new HashMap<>(0))).build();
+    }
+    
+    @RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
+    @GET
+    @Produces(MediaType.TEXT_HTML)
     @Path("/item/{token}")
     public Response viewItemById(@Context HttpServletRequest req, @PathParam("token") String token) throws URISyntaxException {
         Map<String, String> map = new HashMap<>(1);
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
@@ -72,7 +72,7 @@ public class PublishResource {
     @RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
     @GET
     @Produces(MediaType.TEXT_HTML)
-    @Path("/item/raster")
+    @Path("/layer/raster")
     public Response createRasterItem(@Context HttpServletRequest req) throws URISyntaxException {
        return Response.ok(new Viewable("/WEB-INF/jsp/publish/item/raster.jsp", new HashMap<>(0))).build();
     }
--- a/coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
+++ b/coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
@@ -154,7 +154,6 @@
 							</button>
 							<ul class="dropdown-menu" role="menu">
 								<li><a id="publish-button-create-item-option" href="#">Item</a></li>
-								<li><a id="publish-button-create-raster-item-option" href="<%=baseUrl%>/publish/item/raster">Raster Item</a></li>
 								<li><a id="publish-button-create-aggregation-option" href="#">Aggregation</a></li>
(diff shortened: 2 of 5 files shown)


coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
```diff
@@ -154,6 +154,7 @@
 							</button>
 							<ul class="dropdown-menu" role="menu">
 								<li><a id="publish-button-create-item-option" href="#">Item</a></li>
+								<li><a id="publish-button-create-raster-item-option" href="<%=baseUrl%>/publish/item/raster">Raster Item</a></li>
```
cschroed-usgs: This permits users to discover the new page.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/raster.jsp
```diff
@@ -0,0 +1,98 @@
+<%@page import="java.util.HashSet"%>
```
cschroed-usgs: None of the javascript, JNDI stuff is used in any meaningful way yet. I may add JS to leverage fineuploader to display a progress spinner, validate on filetype, and do automatic redirects to the edit page on successful upload.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
```diff
@@ -72,6 +72,14 @@ public Response viewBlankItem(@Context HttpServletRequest req) throws URISyntaxE
     @RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
     @GET
     @Produces(MediaType.TEXT_HTML)
+    @Path("/item/raster")
```
jiwalker-usgs: It might be best to make this path just `/raster` by going with `/item/raster` it collides a bit with the path below where the itemId is somehow "raster".  It could be helpful to match the data paths a bit, making this `/publish/layer/raster` so that it is clear that a layer is being created and possibly separate it from the vector layers that will also need a page somewhat like this.  Otherwise, just going with `/publish/layer` with the assumption that one page for the layers is enough would be good.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
jiwalker-usgs: It might be better to create this link somewhere else, this ends up adding to the confusion of items vs. layers.  Now that we have two layer types (vector and raster) and four item types (uber, aggregate, template, data) we should make it more clear how these operate.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/raster.jsp
jiwalker-usgs: I'm not sure how the redirect will work (I think it is a good idea though), as we need some transfer of the layer id that is created.  It might be good to add a new entry point to that page where it is creating a new item from an existing layer and that could be put in the path.  Something like `/publish/item/fromlayer/{id}` if I'm just spit-balling.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
cschroed-usgs: Ok. I will move it.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/raster.jsp
cschroed-usgs: I like that idea

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
cschroed-usgs: Good suggestions. I like `/publish/layer/raster` most of all.

cschroed-usgs:
Thanks for your feedback @jiwalker-usgs . Sorry I didn't respond to it sooner. I had bad 3 year old email filters that hid notifications for this repo. The filters are fixed now.

coastal-hazards-portal/src/main/webapp/js/application/publish/raster.js
```diff
@@ -0,0 +1,46 @@
...
+			$result.empty();
+			
+			var status = jqXHR.status;
+			var layerUrl = jqXHR.getResponseHeader('Location');
+			var layerId = getLayerIdFromUrl(layerUrl);
+			if(201 === status){
+				
+				$result.append("Successfully published layer . Click ");
+				$result.append('<a href="' + layerUrl + '">here</a> to see the layer');
+
```
cschroed-usgs: I need to better understand the existing process for creating a vector layer before I can provide better functionality.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/raster.jsp
cschroed-usgs: Now that I have touched this up and tested it against the new service code, I understand what you mean better. I originally neglected that layers != items so I thought that I would be able to send users to `/publish/item/{layer-id}`. I could use more suggestions on how to make something like `/publish/item/fromlayer/{id}`. I would also like to understand how we currently create items after the creation of a vector layer.

cschroed-usgs: Megan filled me in on the status quo. It sounds like we currently use some R scripts to create vector layers and items. Since this functionality isn't present in the web UI for vector data, we're probably going to punt on something like the `/publish/item/fromlayer/{id}` and revisit it in a subsequent ticket.

PR #1498, opened 2016-11-08
Cchs 258

cschroed-usgs:
This pull request changes the way that we determine the color swatch in the identify balloon.

Prior to this work, the status quo for determining the color swatch in the identify balloon was to first examine the domain value (probability of collision, likelihood of coastal response. etc.) under the click.  Then we picked a bin by iterating through the bins of the item's SLD Info, stopping at the bin that contained the domain value. We used the bin's color property as the color for the swatch in the identify balloon.

This pull request eliminates that approach in favor of using the color of the layer's pixel underneath the click. This was manually tested when 0, 1, or more raster layers were active.

It is still useful for the client to request the SLDInfo for the binned items, because it still needs information on how to map the domain value under the click to a textual category.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
@@ -59,6 +59,42 @@ CCH.Objects.LayerIdentifyControl = OpenLayers.Class(OpenLayers.Control.WMSGetFea
 
 		return color;
 	};
+	
+	/**
+	 * @param {Number} x the 'x' coordinate of the desired pixel
+	 * @param {Number} y the 'y' coordinate of the desired pixel
+	 * @param {2dCanvasContext} canvasContext
+	 * @returns {String} valid css color for the pixel beneath the click event
+	 */
+	var getCanvasPixelColor = function(x, y, canvasContext){
+		// https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData
+		// html canvas context api for getImageData expects following order of args:
+		// (x, y, width, height)
+		var rgba = canvasContext.getImageData(x, y, 1, 1).data;
+		var strRgba = "rgba(" + rgba.join(rgba, ',') + ")";
+		return strRgba;
+	};
+	
+	/**
+	 * 
+	 * @param {OpenLayers.Pixel} xy
+	 * @param {OpenLayers.Map} map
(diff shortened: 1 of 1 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -84,12 +89,18 @@ return {
...
 			splitName,
 			layerId,
-			trimLayerName = function (name) {
+			/**
+			 * Trims extra information from the layer name so that
+			 * only the item id remains
+			 * @param {String} layerName layer name
+			 * @returns {String} item id
+			 */
+			trimLayerName = function (layerName) {
```
cschroed-usgs: The name of this function confused me. It looks like it is actually getting the item id from a GeoServer WMS layer name. Is it reasonable to change this to 'getItemNameFromLayerName'? Or am I misunderstanding, and it is cool as-is?

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -119,6 +129,8 @@ return {
 			// over with a different SLD for each layer. In order to handle
 			// that, I need to make an array for the layer name (item.id) 
 			// to be able to process this going forward
+			var layerNameToPixelColor = {};
```
cschroed-usgs: If I change the name of the `trimLayerName` function, would it make sense to change this too?

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -119,6 +129,8 @@ return {
...
+			
 			cchLayers.each(function (l) {
 				var lName = trimLayerName(l.name);
```
cschroed-usgs: ... and this var to `itemId`?

jiwalker-usgs: I think it would probably be useful to bring @isuftin in on these questions.

cschroed-usgs:
@jiwalker-usgs I'm not sure. I don't know how to load ribbons locally. I will try to do that.

cschroed-usgs:
Megan said it's cool to merge this and test against the broad range of datasets on dev vs. trying to get them imported in my local environment.

PR #1505, opened 2016-11-17
CCHS-260 PAE updates to display correct colors

smlarson-usgs:


smlarson-usgs:
Carl and I reviewed this change and he is fine with the merge.

PR #1507, opened 2016-11-21
Removing defunct repository

cschroed-usgs:


pom.xml
```diff
@@ -157,11 +157,6 @@
 			<url>http://download.osgeo.org/webdav/geotools/</url>
 		</repository>
 		<repository>
```
cschroed-usgs: @smlarson-usgs : it turns out we don't need this repo any more. Pulling this fix will increase your build speed since it will cease to retry dependency retrieval against a hung maven repo.

smlarson-usgs: Good find! thanks

cschroed-usgs: de nada!

PR #1509, opened 2016-11-23
CCHS-225-touch-ups

cschroed-usgs:
The changes here permit a user to create a raster layer in CCH via curl or via the web form. After uploading the user can visit the URL for the new CCH layer object . The layer object contains a 'serviceParameter' property. If you copy the value of this property, you can search for it in the GeoServer layer preview list, verify that it exists, and explore the dataset with OpenLayers. UPDATE: I have taken it to the next step -- I have created an item that relies on the raster layer and interacted with it in the portal web map ui.

To troubleshoot the raster layer import I had to perform a bunch of manual iterative tests. I couldn't find a way to quickly rename a TIFF and its zip file without re-zipping the whole huge TIFF each time (holy slow roller batman!) so I needed to find a way to upload the same file multiple times. I changed the TIFF filenames and the GeoServer layers (not to be confused with the CCH Layer objects) to use random ids.instead of the original filename so that I could upload the same zipped tiff multiple times. To simplify the implementation, the random ids for the TIFF filenames and the GeoServer layers are unrelated to the CCH layer ids (ex: ESF8pnhi). If desired, we could make them the same by supporting a file name parameter on the FetchAndUnzip WPS Process (1-pointer).

As I troubleshootered (or however that should be said;) ), I found some code problems, some configuration problems, and some data problems that were preventing the raster data from being exposed as layers in GeoServer. It is possible that we don't /need/ the random ids for the Geoserver layers+TIFF filenames in order for the system to function but at this point it may be annoying to back out those changes in tandem with the other necessary ones.

Code changes:
* UUID stuff, thus eliminating the need for raster files to have distinct names.
* Removed bug that caused files with multiple "."s in their name from being considered the same. Previously, 'ne_ae2020.carl.zip' and 'ne_ae2020.zip' were both being created as 'ne_ae2020' and thus failing.
* Converted a silent failure to a loud one when a raster store+layer could not be created on GeoServer. Even if we back out on UUIDs, this is especially important to keep.
* Added performance time logging for long-running operations to gain additional insight into bottlenecks and whether the apps were even still alive during testing ;)
* Parameterized a WMSService creation helper method so that custom workspaces could be specified. 

In the process of pursuing this last item, I looked through other code and noticed that the proxied workspace is often not parameterized ([example](https://github.com/USGS-CIDA/coastal-hazards/pull/1509/files#diff-a8c0cbcc67534bd2802a9b48ec2ba044R229)). Is workspace separation important enough to warrant parameterizing the workspace on several methods (or many? Sorry, I don't know how many yet), and updating usages accordingly?
@mhines-usgs, @smlarson-usgs: Do we really need to be putting raster files in their own dedicated workspace? In the short term it would likely be simpler to put both raster and vector in the same workspace. In the long term, I don't yet understand what we gain from a workspace separation. Could one of you please elaborate on the advantages of segregating the raster and vector data into different workspaces from a data management perspective?

Necessary local configuration changes:
* I ensured that the the context.xml's for both the geoserver instance and the portal instance had vars for the shared "fetch and unzip" token
* I ensured that the context.xml for the geoserver instance had a var for the "fetch and unzip" base dir.
* I ensured that the self-signed certs being presented by my local tomcat instances were both trusted in the system Java keystore. Previously only one of the certs was trusted.
* I synced my GeoServer admin password and the coastal-hazards.geoserver.password context.xml variable.
^ One or more of these may need to be performed on dev.

Data changes:
* I logged into GeoServer and created the "portal" workspace. I did this on dev too.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
@@ -146,7 +146,7 @@ public class LayerResource {
         @POST
 	@Path("/raster")
 	@Consumes(MediaType.MULTIPART_FORM_DATA)
-	@Produces(MediaType.APPLICATION_JSON)
+	@Produces(MediaType.TEXT_PLAIN)
 	@RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
 	public Response createRasterLayer(
                 @Context HttpServletRequest req, 
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
@@ -146,7 +146,7 @@ public class LayerResource {
         @POST
 	@Path("/raster")
 	@Consumes(MediaType.MULTIPART_FORM_DATA)
-	@Produces(MediaType.TEXT_PLAIN)
+	@Produces(MediaType.APPLICATION_JSON)
 	@RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
 	public Response createRasterLayer(
                 @Context HttpServletRequest req, 
@@ -175,9 +175,14 @@ public class LayerResource {
                 services.add(MetadataUtil.makeCSWServiceForUrl(MetadataUtil.getMetadataByIdUrl(metadataId)));
 
                 Bbox bbox = MetadataUtil.getBoundingBoxFromFgdcMetadata(metadata);
+                log.info("Starting CRS Identifier Lookup");
+                long startTime = System.nanoTime();
                 String EPSGcode = CRS.lookupIdentifier(MetadataUtil.getCrsFromFgdcMetadata(metadata), true);
+                long endTime = System.nanoTime();
+                long duration = (endTime - startTime)/1000000;  //divide by 1000000 to get milliseconds
+                log.info("Finished CRS Identifier Lookup. Took " + duration + "ms.");
                 
-                if (bbox == null || EPSGcode == null) {
+               if (bbox == null || EPSGcode == null) {
                     throw new ServerErrorException("Unable to identify bbox or epsg code from metadata.", Status.INTERNAL_SERVER_ERROR);
                 }
                 
@@ -186,8 +191,11 @@ public class LayerResource {
                 log.info("Raster layer create - about to addRasterLayer to geoserver with an EPSG of: " + EPSGcode);
                 
                 Service rasterService = GeoserverUtil.addRasterLayer(geoserverEndpoint, zipFileStream, newId, bbox, EPSGcode);
-               
-                services.add(rasterService);
+                if(null == rasterService) {
+                        throw new ServerErrorException("Unable to create a store and/or layer in GeoServer.", Status.INTERNAL_SERVER_ERROR);
+                } else {
+                        services.add(rasterService);
+                }
                 
                 if (!services.isEmpty()) {
 			Layer layer = new Layer();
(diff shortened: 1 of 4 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
```diff
@@ -186,8 +191,11 @@ public Response createRasterLayer(
                 log.info("Raster layer create - about to addRasterLayer to geoserver with an EPSG of: " + EPSGcode);
                 
                 Service rasterService = GeoserverUtil.addRasterLayer(geoserverEndpoint, zipFileStream, newId, bbox, EPSGcode);
-               
-                services.add(rasterService);
+                if(null == rasterService) {
+                        throw new ServerErrorException("Unable to create a store and/or layer in GeoServer.", Status.INTERNAL_SERVER_ERROR);
```
cschroed-usgs: This error handling was lacking before, thus masking errors happening via GeoServer. I'm open to suggestion on whether this should be implemented as it is currently (a null-check-on-return), or as a thrown-and-caught exception.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -365,13 +366,17 @@ public static Service addRasterLayer(String geoServerEndpoint, InputStream zipFi
...
-                    log.info("Published GeoTiff!!!");
-                    log.info("In GeoserverUtil, about to add wmsService with layer name: " + layerId);
-                    Service rasterService = wmsService(layerId);
-                    log.info("Added layer to wms service.");
-                    
+                    RESTCoverageStore store = publisher.publishExternalGeoTIFF(PORTAL_WORKSPACE, fileName, unzippedFile, coverageEncoder, layerEncoder); // #TODO# what to do if the workspace:fileName is already in use?
+                    if (null == store) {
+                            //if store or layer creation failed
+                            log.info("Error publishing GeoTiff in GeoServer.");
+                            rasterService = null;
```
cschroed-usgs: Here we are returning null. We could also throw an Exception instead.

coastal-hazards-wps/src/main/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcess.java
```diff
@@ -114,23 +116,28 @@ ZipInputStream getZipFromUrl(String zipUrl, HttpClient client){
         return zipStream;
     }
     
-    String unzipToDir(ZipInputStream zipStream, File zipDir) {
```
cschroed-usgs: Changed from String to File to make it easier to test.

coastal-hazards-wps/src/test/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcessTest.java
```diff
@@ -135,45 +138,24 @@ public void testGetZipFromGoodUrls() {
         }
         return; //No Exceptions were thrown on any HTTP status codes
```
cschroed-usgs: Previously this file tested only whether the the expected file names and subdir names existed after unzipping. The previous tests ignored the contents of the files. The new tests ignore the expected file name and instead validate that the unziped file exists, is readable, and has the expected contents.

cschroed-usgs: TODO: see if we can remove some of the old wps zip test resources. Some are no longer needed because the process only handles single-file zips now.

cschroed-usgs:
If testing this via curl, ensure that the -v flag is used so that you can see the response headers. The 'Location' response header is the url you would visit to see the JSON for the new CCH Layer object.

coastal-hazards-wps/src/test/java/gov/usgs/cida/coastalhazards/wps/FetchAndUnzipProcessTest.java
smlarson-usgs: I believe a Jira ticket was going to be created for what the system would do when it detected a file that already exists. It was recommended to proceed with the code release without having this requirement at the time. You may want to see if that ticket should be closed.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -78,7 +79,7 @@
 
 			if (StringUtils.isNotBlank(created)) {
 				serviceList.add(wfsService(created));
-				serviceList.add(wmsService(created));
+				serviceList.add(wmsService(created, PROXY_WORKSPACE));
```
smlarson-usgs: Jordan suggested that the two workspaces will need to be supported for now. There are unknown dependencies that would have to discovered and removed prior to converting down to just one workspace.

PR #1515, opened 2016-12-06
Automated addition of logicalFilePath to all databaseChangeLog elements.

cschroed-usgs:
Is this a good approach to resolve the logical file path mismatches in the qa changelog? Note that we are targeting the hotfix branch with this pull request, not master.


cschroed-usgs:
Introduced these changes by creating a script addLogicalFilePath.sh:
```
ONLY_FILE_NAME=`basename ${1}`;
sed -i \
"s/<databaseChangeLog/<databaseChangeLog logicalFilePath=\"liquibaseChangeSet\/${ONLY_FILE_NAME}\"/" ${1};
```
then...
```
cd coastal-hazards-liquibase
find . -iname '*.xml' -exec bash addLogicalFilePath.sh {} \;
```

jiwalker-usgs:
We should double check that this matches the table on prod.  I think the previous way of doing this was a bit more manual, so there could be typos or mismatches.

jiwalker-usgs:
Other than that I think it looks fairly solid.

cschroed-usgs:
To verify that my auto changes didn't break any xml syntax rules I created `isValidXml.sh`:
```
xmllint --noout ${1}
#output xmllint's exit code and the filename
echo "$?:${1}"
```
Then from `coastal-hazards-liquibase` I ran:
```
find . -iname '*.xml' -exec bash isValidXml.sh {} \;
```
I got back all-zero exit codes next to my files. I verified that this was a good test by trying the same command after creating a `bogus.xml` file with bogus contents, and got a '1'-valued exit code.

PR #1517, opened 2016-12-07
Removing some of the quirks of services section

jiwalker-usgs:
* remove required fields
* remove dropdowns for import
* remove import
* add populate from layer
* move booleans to same spot on page

I've tested to the point of actually doing anything with these, I will need to load some real data to see what happens here when buttons are used.  Likely more will be required here to finish this off.  I included Ivan in review as he is an interested party, but don't feel obligated to dive in too far.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
@@ -71,6 +71,8 @@ CCH.Objects.Publish.UI = function () {
 		$proxyWfsCheckButton = $form.find('#form-publish-item-service-proxy-wfs-import-button-check'),
 		$proxyWmsCheckButton = $form.find('#form-publish-item-service-proxy-wms-import-button-check'),
 		$getWfsAttributesButton = $form.find('#form-publish-item-service-proxy-wfs-pull-attributes-button'),
+		$popFromLayerInput = $form.find('#form-publish-item-service-layer'),
+		$popFromLayerButton = $form.find('#form-publish-item-service-layer-button-pop'),
 		$emphasisItemSpan = $form.find('.emphasis-item'),
 		$emphasisAggregationSpan = $form.find('.emphasis-aggregation'),
 		$isActiveStormRow = $form.find('#form-publish-info-item-active-storm'),
@@ -155,11 +157,12 @@ CCH.Objects.Publish.UI = function () {
 		
 		[$titleFullTextArea, $titleMediumTextArea, $descriptionFullTextArea,
 			$descriptionMediumTextArea, $descriptionTinyTextArea, $bboxNorth,
-			$typeSb, $attributeSelect,
-			$srcWfsServiceInput, $srcWfsServiceParamInput,
+			$typeSb, $attributeSelect, $isFeaturedCB, $ribbonableCb,
+			$popFromLayerInput, $popFromLayerButton,
+			$cswServiceInput, $srcWfsServiceInput, $srcWfsServiceParamInput,
 			$srcWmsServiceInput, $srcWmsServiceParamInput, $proxyWfsServiceInput,
 			$proxyWfsServiceParamInput, $proxyWmsServiceInput, $getWfsAttributesButton,
-			$proxyWmsServiceParamInput, $ribbonableCb, $name, $wfsServerHelpButton,
+			$proxyWmsServiceParamInput,  $name, $wfsServerHelpButton,
 			$wfsSourceCopyButton, $sourceWfsCheckButton,
 			$sourceWmsCheckButton, $wmsServerHelpButton, $proxyWfsCheckButton,
 			$proxyWmsCheckButton, $buttonSave, $buttonDelete,
@@ -170,8 +173,6 @@ CCH.Objects.Publish.UI = function () {
 					$item.removeAttr(CCH.CONFIG.strings.disabled);
 				});
 		
-		$proxyWfsServiceInput.val(gsBaseUrl + 'wfs');
-		$proxyWmsServiceInput.val(gsBaseUrl + 'wms');
 		$showChildrenCb.prop(CCH.CONFIG.strings.checked, false);
 		$isActiveStormChecbox.prop(CCH.CONFIG.strings.checked, false);
 		$isFeaturedCB.prop(CCH.CONFIG.strings.checked, false);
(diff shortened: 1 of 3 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -263,51 +264,33 @@ CCH.Objects.Publish.UI = function () {
 				}
 				
 
-				if (me.isBlank($srcWfsServiceInput)) {
```
cschroed-usgs: What do you think about DRYing up this section up a bit? I can think of a few ways. One way would be to make an object literal that maps String parameter names (ex: "WFS Source Endpoint", "WMS Source parameter") to their corresponding jQuery objects (ex: "$srcWfsServiceInput", "$srcWmsServiceParamInput"). Then you could iterate over the key-value pairs, validating the length using the iteration's current *value*, and then dynamically construct an error message using the iteration's current *key*.

cschroed-usgs: If we remove these calls to `me.isBlank`, are there other usages of `me.isBlank`. Could it be removed?

jiwalker-usgs: I believe there are other usages of me.isBlank

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -1369,6 +1352,47 @@ CCH.Objects.Publish.UI = function () {
...
 	};
+	
+	me.loadLayerInfo = function (layerid) {
+		if (layerid) {
+			var layerurl = CCH.CONFIG.contextPath + '/data/layer/' + layerid;
+			$.ajax({
+				url: layerurl,
+				success: function (data) {
+					for (var service in data.services) {
+						if (service.type === "csw") {
```
jiwalker-usgs: btw, I thought about this in a similar vein with the DRYing, but opted for this.  let me know if you disagree.

cschroed-usgs: I tend to prefer to convert predictably-structured conditionals like this into object literals. I like doing that because it converts the procedural association of primitives to objects (or functions) into a declarative association. The technique reliably reduces the number of procedural lines. It also often reduces the overall number of lines. What do you think of that technique in general? Do you think it applies in this particular situation?

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -1369,6 +1352,47 @@ CCH.Objects.Publish.UI = function () {
...
 		}
 	};
+	
+	me.loadLayerInfo = function (layerid) {
+		if (layerid) {
+			var layerurl = CCH.CONFIG.contextPath + '/data/layer/' + layerid;
+			$.ajax({
+				url: layerurl,
+				success: function (data) {
+					for (var service in data.services) {
```
cschroed-usgs: Douglas Crockford and I aren't fans of `for in` loops.
Crockford wrote about this in JavaScript: The Good Parts back in the ECMAScript 3 days. There weren't very good alternatives then.
I expanded upon Crockford's ideas more recently, detailing improved alternatives in native ECMAScript 5 and in common JS utility libraries:
https://www.carlschroedl.com/blog/comp/avoid-for-in-in-javascript/2014/11/05/

Do you understand the risks associated with `for in` loops? Do any of these alternatives appeal to you? If you are still a big fan of `for in`, and aren't interested in pursuing alternatives, could you please wrap the body of this loop in a conditional that checks `object.hasOwnProperty`?.

jiwalker-usgs: Yeah, I'll change this.  Switching between R and JavaScript has me in some bad habits.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
cschroed-usgs: This is a new function. Are there side effects to DRYing a new function that would expand the scope of the ticket? While I disagree with the assessment of scope creep on DRYing the other existing method for validating the length of endpoints and parameters, I can think of abstract reasons why a developer would resist DRYing an existing method, and trust in your superior knowledge of the mediation page's context to make the call. I'm having a harder time understanding why a developer would resist low-hanging DRY fruit on new code**. Could you please help me understand in general or in this particular case? What do you think of the object literal technique that I mentioned in general? Do you think it applies here?

**Unless it were literally a low-hanging dry fruit... prunes are the worst ;)

cschroed-usgs: I was thinking of something like this.

Outside the loop...
```
var stupendousMap = {
'source_wfs' : [$srcWfsServiceInput, $srcWfsServiceParamInput],
...
};
```
Inside the loop...
```
var setVals = function($serviceInput, $paramInput){
   $serviceInput.val(service.endpoint);
   $paramInput.val(service.serviceParameter);
}
var inputAndParamInput = stupendousMap.getOwnProperty('type');
setVals.apply(inputAndParamInput);
```

cschroed-usgs: I just noticed that the 'csw' type requires different behavior from the rest of the types, which makes the approach I advocated for less attractive.

cschroed-usgs: I can also buy the argument that it's a weak contract to implicitly require identical ordering in the array values in `stupendousMap` and in `setVal`'s function arguments.

cschroed-usgs: Sorry Walker, there are more tradeoffs here than I originally perceived. I would do it a different way, but your approach is reasonable. Differences in style are ok and coolio.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
cschroed-usgs: No worries. The Struggle Is Real (TM)(C). I'm glad we are on the same page w.r.t. JS's broken-by-default-ness :).

cschroed-usgs:
After you take a crack at making the object iteration a bit safer, this will be good to go by me!

cschroed-usgs:
I like the idea of grouping the booleans together. :+1: to making the populate from layer metadata functionality more explicit than the current dropdown. For that, are you thinking of going for a modal dialog with a search thing on it? <-- just curious.

jiwalker-usgs:
We'd have to expose a search service, I was just thinking copy the id from the result of an upload and paste that into the box.

cschroed-usgs:
Good thinking!

jiwalker-usgs:
Fixed for in loop and added some stuff.  Testing vector locally again, but I want to test on dev before I call this good enough.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -1359,21 +1370,24 @@ CCH.Objects.Publish.UI = function () {
 			$.ajax({
 				url: layerurl,
 				success: function (data) {
-					for (var service in data.services) {
+					for (var i=0; i < data.services.length; i++) {
```
cschroed-usgs: Oh I thought we were iterating over an object's members, not an array of object! Looking good.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -1359,21 +1370,24 @@ CCH.Objects.Publish.UI = function () {
...
+						var service = data.services[i];
+						var serviceEndpoint = (service.hasOwnProperty("endpoint")) ? service.endpoint : "";
```
cschroed-usgs: While it probably doesn't hurt, I'm curious: is there a need for the hasOwnProperty check here? hasOwnProperty is essential if using `for in` loops on objects because it can be used as a condition for processing the body of a loop based on whether the object key of the current iteration refers to a property on the actual object, or a property belonging to an object further up its prototype chain. I'm not sure it's essential here. What do you think? This isn't a blocker, just trying to learnify.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -753,13 +758,19 @@ CCH.Objects.Publish.UI = function () {
 					$option = $('<option>')
 							.attr('value', ft.name)
 							.html(ft.name);
-					$attributeSelect.append($option);
+					$attributeSelectHelper.append($option);
 				}
 			});
 		}
-		$attributeSelect.removeAttr(CCH.CONFIG.strings.disabled);
```
cschroed-usgs: Never mind, I noticed it's disabled in the html https://github.com/USGS-CIDA/coastal-hazards/pull/1517/files/9d5f54083da686c9825994858415dbaefec15c41..d9581fbdebdafabe8987613cd0ece3ad3a048156#diff-6dc2791c54b5452c8563d0b756b40c4cR416

cschroed-usgs:
No  blockers here. Merge when ready.

jiwalker-usgs:
Thanks @cschroed-usgs

PR #1518, opened 2016-12-08
changing service_endpoint column from character varying(255) to text

mhines-usgs:
ran into an issue with changing the service_endpoint column contents to use https, because some URLs are actually 255 characters right now and can't fit in the s on https


mhines-usgs:
@jiwalker-usgs & @cschroed-usgs not sure if there are other implications of this change especially to the column type. If you'd prefer I just raise the character limit, I can change this changeset

cschroed-usgs:
Running
```
SELECT MAX(LENGTH(service_parameter)) FROM service;
```
on qa yields '89'.
on prod yields '65'.
We're probably good to keep service_parameter's column type as-is.

mhines-usgs:
Nice fast investigation :) I was just going to do the same.

mhines-usgs:
RE: my question about other changes we might need, are probably unnecessary. I searched the repository for service_endpoint and found nothing particularly alarming.

cschroed-usgs:
Good find @mhines-usgs. That could cause the app to error. I'm not familiar enough with the `@Column` annotation to recommend what to change. @jiwalker-usgs ?

jiwalker-usgs:
I've run into this in the past, and what the internet says is that text is basically better than varchar in every way except compatibility.  I don't think the app will fail with that length in the column annotation, though it might as well be removed since we don't have a limitation anymore.

mhines-usgs:
@jiwalker-usgs or @cschroed-usgs can you look over the small hibernate change here? Thanks :D

mhines-usgs:
https://github.com/USGS-CIDA/coastal-hazards/pull/1518/commits/f90d261de7ebd05ec5bd2ee06924fbdde5115cef

cschroed-usgs:
Approv'd

PR #1519, opened 2016-12-08
Reverting url change to maintain backward compatibility.

cschroed-usgs:
Smallest PR evar. I'm hoping this backwards compatibility makes our lives much easier as we juggle the raster and https work across tiers.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/LayerResource.java
@@ -95,7 +95,7 @@ public class LayerResource {
 	}
 	
 	@POST
-	@Path("/vector")
+	@Path("/")
 	@Consumes(MediaType.APPLICATION_OCTET_STREAM)
 	@Produces(MediaType.TEXT_PLAIN)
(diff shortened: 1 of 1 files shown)


cschroed-usgs:
This change eliminates the concerns I expressed in [a JIRA comment|https://internal.cida.usgs.gov/jira/browse/CCHS-283?focusedCommentId=117940&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-117940]:

```
'/vector' is added to the URL only when you are POSTing, so only when you are creating a layer. GETs/request for layers should remain the same.
...
Also, that change in the web service url only applies to dev for sure because I think that is the only place we have deployed those raster changes. We may or may not have deployed the path change to the layer creation service on qa. We have definitely not deployed the path change to the layer creation service on prod.
```

cschroed-usgs:
Tests all pass after this change.The following greps from my history don't reveal anything worrisome:
```
 1010  grep -irl '/vector' . | grep -v '/target/'
 1011  grep -irl 'vector' . | grep -v '/target/'
 1012  grep -irl '"vector' . | grep -v '/target/'
 1013  grep -irl "'vector" . | grep -v '/target/'
```

cschroed-usgs:
Going to merge. I'm still interested in @jiwalker-usgs 's opinion. We can easily revert this revert later if need be. revert.

cschroed-usgs:
@jiwalker-usgs I like that idea. My Jersey-fu is weak. Do you know of a quick win?

PR #1523, opened 2016-12-14
changing metadataUtil.java to use new external endpoint

mhines-usgs:


jiwalker-usgs:
We'll need to be sure to add this property to context.xml on all tiers.

PR #1524, opened 2016-12-16
CCHS-280: mediation page

jiwalker-usgs:
* select metadata or csw endpoint can populate keywords and bbox
* attribute now shows up when editing item
* populate resources button is enabled
* removed limit on service endpoints

The rest of the work for this should relate to hazardItems package

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Service.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Service.java
@@ -1,7 +1,5 @@
 package gov.usgs.cida.coastalhazards.model;
 
-import gov.usgs.cida.coastalhazards.model.summary.Full;
-import gov.usgs.cida.coastalhazards.model.summary.Summary;
 import gov.usgs.cida.coastalhazards.util.ogc.CSWService;
 import gov.usgs.cida.coastalhazards.util.ogc.OGCService;
 import gov.usgs.cida.coastalhazards.util.ogc.WFSService;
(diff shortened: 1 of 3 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -830,7 +833,7 @@ CCH.Objects.Publish.UI = function () {
 				$(evt.target).closest('.well').remove();
 			});
 			$rowObject.find('select').val(type);
-			$rowObject.find('select').on('change', me.resourceTypeChanged);
+			$rowObject.find('select').on(CCH.CONFIG.strings.change, me.resourceTypeChanged);
```
cschroed-usgs: 👍 for favoring programmatic identifiers over strings !

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
```diff
@@ -300,7 +300,12 @@
 									<div id="form-publish-info-item-service-csw" class="row row-csw">
 										<div class="form-group">
 											<label for="form-publish-item-service-csw">CSW</label>
-											<input type="text" class="form-control" id="form-publish-item-service-csw" disabled="disabled" maxlength="<%= Service.ENDPOINT_MAX_LENGTH%>" />
+											<div class="input-group">
+												<input type="text" class="form-control" id="form-publish-item-service-csw" disabled="disabled" maxlength="<%= Service.ENDPOINT_MAX_LENGTH%>" />
```
jiwalker-usgs: Yeah, I thought about removing this, but opted for leaving it how it was. The maxlength will not do anything here, but shouldn't be an issue either.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Service.java
```diff
@@ -28,7 +26,7 @@
 
 	private static final long serialVersionUID = 1L;
 
-	public static final int ENDPOINT_MAX_LENGTH = 255;
+	public static final int ENDPOINT_MAX_LENGTH = Integer.MAX_VALUE;
```
jiwalker-usgs: I thought it would be cleaner to change this rather than remove the precondition check.

PR #1526, opened 2016-12-20
CCHS-293

jiwalker-usgs:
Layers need namespaced names to apply sld correctly.


mhines-usgs:
:+1:

cschroed-usgs:
Nice catch guys!

PR #1528, opened 2016-12-22
Enable caching in proxy in front of GeoServer 

cschroed-usgs:
Enabling GeoServer tiles to be cached by specifying appropriate server response headers to ignore

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/WEB-INF/web.xml
+++ b/coastal-hazards-portal/src/main/webapp/WEB-INF/web.xml
@@ -44,6 +44,10 @@
 			<param-value>coastal-hazards.portal.geoserver.cache.name</param-value>
 		</init-param>
 		<init-param>
+			<param-name>ignored-server-response-headers</param-name>
+			<param-value>expires,cache-control,age</param-value>
+		</init-param>
+		<init-param>
 			<param-name>max-megabytes-on-disk</param-name>
 			<param-value>8096</param-value>
 		</init-param>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -277,7 +277,7 @@
 			<dependency>
 				<artifactId>proxy-utils</artifactId>
 				<groupId>gov.usgs.cida</groupId>
-				<version>1.0.6</version>
+				<version>1.0.6.1</version>
```
cschroed-usgs: This has been released

coastal-hazards-portal/src/main/webapp/WEB-INF/web.xml
```diff
@@ -44,6 +44,10 @@
 			<param-value>coastal-hazards.portal.geoserver.cache.name</param-value>
 		</init-param>
 		<init-param>
+			<param-name>ignored-server-response-headers</param-name>
+			<param-value>expires,cache-control,age</param-value>
```
cschroed-usgs: These were the only cache-related headers that curl WMS GetMap requests were receiving.

cschroed-usgs:
How will this interact with the real time storm functionality?

cschroed-usgs:
Should we consider setting up a separate CachingProxyServlet and url mapping exclusively for raster layers? Would opting in to a raster-only cache necessitate a simple update of the urls in the db, or would we need to change how urls were built throughout the app ?

cschroed-usgs:
Is there a pre-GetMap request in OpenLayers that we could use to catch the current zoom level/bounding box, and then redirect the GetMap request to a CachingProxyServlet url that was exclusively for far-zoomed-out tiles?

cschroed-usgs:
Never mind my questions. I ignored the fact that caching is opt-in on a layer-by-layer basis when I was writing those questions.

PR #1529, opened 2016-12-22
Cchs 301

cschroed-usgs:
Each tile retrieval is timed. The entire script can be timed by wrapping the script in a call to `time`:
```
time bash rasterCurls.sh https://cida-test.er.usgs.gov/dev/coastalchangehazardsportal _0b57c74f-d8d6-4686-a4fb-0de9b5c7f6d0 EWyGbdyH
```


seedAllRasters.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
cschroed-usgs: entry point

jiwalker-usgs:
I think this looks fine, my one comment would be to move them into one of the modules (portal or geoserver probably).  Having them at the root level seems slightly out of place.

jiwalker-usgs:
Also, a comment for future improvement (if this ends up being a longer term solution) would be to only go by the item_id and add a step where it gets the wms url from the services section of that item.

cschroed-usgs:
Thanks @jiwalker-usgs . I like your idea for improvement and I will move the scripts into a module.

coastal-hazards-portal/invalidateRasterCache.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
cschroed-usgs: Suzanne had the good idea of splitting invalidation out into a separate step. I will create a separate Jenkins job for seeding vs invalidating. I can create a parent job that calls both.

PR #1532, opened 2016-12-28
removing host and referrer which were causing waf problems

mhines-usgs:
getting 404s on production when running this job. Tested without referrer by curling and got WAF response, so removed host and referrer and the PNGs are returned as desired.


cschroed-usgs:
👍

PR #1534, opened 2017-01-03
Improving docker docs

cschroed-usgs:


mhines-usgs:
@cschroed-usgs who do you think should be peer reviewing this work?

cschroed-usgs:
I welcome review from either you or Suzanne. Since you had a big part in constructing the Google doc I think it would be helpful for you to confirm that any useful tidbits that I am proposing to eliminate from the Google doc are present in this README.

Docker/README.md
```diff
@@ -37,6 +38,17 @@ If you are building these containers on a machine on the DOI network, there is a
...
 With this modification, the containers that need it will include pulling the SSL root certificate from DOI and install it into openssl as a valid certificate.
 
+#### Parameterized Environment Variables
+===============================
+Several of the containers accept parameterized environment files. By default, `docker-compose` will use `compose.env`. To customize the parameters, first create a copy of `compose.env`. Make sure the copy's file name starts with`compose` and ends with `.env`. An example valid custom env file name is `compose_johns_laptop.env`. To take advantage of your custom env file, prepend each of your `docker-compose` commands with an assignment to the `CCH_ENV_LOCAL` variable. Use the middle portion of your custom env file name as the value. For example:
+
+```
+$ CCH_ENV_LOCAL="_johns_laptop" doi_network="true" docker-compose up cch_db cch_rserve cch_n52_wps cch_pycsw
+```
+
```
smlarson-usgs: Consider making a naming convention for all local envs ...doi_network="true" CCH_ENV_LOCAL="_local" docker-compose build  <it may make more sense when interpreting the directions later>. Also, an example file outlining what is different locally with a particular set-up would be very valuable.

Docker/README.md
```diff
@@ -78,9 +90,38 @@ Once the server is up and running, you can access the server by pointing your we
...
+This container is standalone and does not depend on any other containers to run. You can run the container by issuing:
+
+`docker-compose up cch_db`
+
+The DOI SSL issue also applies to this container, so if you are on the DOI network, issue:
+
+`doi_network="true" docker-compose build cch_db`
+
+You will then be able to connect to the database via the ports and credentials described in `docker-compose.yml` and `postgres/Dockerfile`
```
smlarson-usgs: An example file that would use this and the components involved in the configuration would be helpful (ie change the context.xml for the portal when you set up these components locally: <list of candidate components>.

Docker/README.md
```diff
@@ -78,9 +90,38 @@ Once the server is up and running, you can access the server by pointing your we
...
+PyCSW requires a database. Accordingly, this container requires the postgres (`cch_db`) container.
+
+You can run the container and its dependent db container by issuing:
+
+`docker-compose up cch_pycsw`
+
+The DOI SSL issue also applies to this container, so if you are on the DOI network, issue:
+
+`doi_network="true" docker-compose build cch_pycsw`
+
```
smlarson-usgs: Struggling with the context again...what components are intended to run locally, which ones are in the Docker containers, ...

smlarson-usgs: hmmm...wondering if we should just supply directions for a specific environment that covers most dev scenarios: all components in Docker containers except for.... and then as needed, the doc could be expanded to avoid creating a matrix of examples that would never be used.

smlarson-usgs: How confident are you that all the Docker components are now colocated and working without the need for modification? If they require mods, what are they? The doc should be tested with a scorched dev env if we want to be really thorough.

cschroed-usgs: > Struggling with the context again...what components are intended to run locally, which ones are in the Docker containers, ...
All the components are available to be run as docker containers. I think it's best to leave it up to individual developers to decide whether they run them locally or in containers based on the tasks that they are tackling.

cschroed-usgs: > hmmm...wondering if we should just supply directions for a specific environment that covers most dev scenarios: all components in Docker containers except for.... and then as needed, the doc could be expanded to avoid creating a matrix of examples that would never be used.

I think it would be ok to document one common use scenario.

What is working for me at the moment is to run the portal app and geoserver locally, and run the rest of the dependent services in containers. Should we try documenting that?

cschroed-usgs: > How confident are you that all the Docker components are now colocated and working without the need for modification? If they require mods, what are they? The doc should be tested with a scorched dev env if we want to be really thorough.

I'd say I'm 80% confident the Docker setup can work without modification for future devs who want to do the same thing that I'm doing with local portal+geoserver, and containerized other services.

I can try a scorched-earth rebuild and compose up, but if it pushes back too hard I'd like to punt since we only have 1 point allocated for this task, and we aren't certain we will continue using Docker.

mhines-usgs:
Nice, I like the readme approach instead of the google doc for the Docker documentation. That strategy makes much more sense, and think the additions you made are helpful. Thanks!

Docker/README.md
smlarson-usgs: That would be excellent!

cschroed-usgs:
Thanks @mhines-usgs !

Docker/README.md
cschroed-usgs: I'm game for switching to consistent casing on the env vars. I don't want to do those mass find-and-replacements in the same pull request as these changes because it makes it much harder to review changes.

cschroed-usgs: The only differences I have in my custom env file are IP addresses and ports that are specific to my setup (native Docker vs Docker Machine, local Portal and GeoServer tomcat installs). I'm afraid an example would confuse people who are using a different setup. I'm also concerned that the example file would get out of sync with the default file pretty fast. Given that the only differences are ports and IP addresses, do you still think it is useful to include an example?

Docker/README.md
cschroed-usgs: Could you please describe what kind of example file you were thinking of?  context.xml? custom env file? something else?

Docker/README.md
cschroed-usgs: I tried the scorched-earth Docker rebuild. I documented a few gotchas and pushed those commits.

Docker/README.md
smlarson-usgs: I do think its still helpful. Even if it gets out of date, its someplace to start.

Docker/README.md
smlarson-usgs: Any of the files that need to be changed (or excerpts from the files) would be helpful. The context changes, where you placed the local file, what you modified...for example.

Docker/README.md
cschroed-usgs: I added an example custom env file

Docker/README.md
cschroed-usgs: I added example context.xmls for running the portal and GeoServer locally.

PR #1535, opened 2017-01-09
Adding RCurl as a native system dependency

cschroed-usgs:
@jiwalker-usgs - you mentioned that RCurl was probably not an inherent need for the hazardItems package, rather that it was needed by the wrapper. Do you think this is a better place to capture the dependency than the imports in the hazardItems DESCRIPTION file?


cschroed-usgs:
Thanks for reviewing! I switched to the shorter command like you suggested.

PR #1536, opened 2017-01-09
cchs-306 added health check for pycsw using getCapabilities and a def…

smlarson-usgs:
…ault version of 2.0.2

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
@@ -55,7 +55,7 @@ public class HealthResource {
 		geoserverEndpoint = props.getProperty("coastal-hazards.portal.geoserver.endpoint");
 		geoserverUser = props.getProperty("coastal-hazards.geoserver.username");
 		geoserverPass = props.getProperty("coastal-hazards.geoserver.password");
-                pycswVersion = props.getProperty("coastal-hazards.csw.version");
+                pycswVersion = props.getProperty("coastal-hazards.csw.version", pycswVersionDefault);
                 pycswEndpoint = props.getProperty("coastal-hazards.csw.external.endpoint");  // includes the / csw/
 	}
 
@@ -148,13 +148,8 @@ public class HealthResource {
                 try { // health check add for pycsw Jira cchs-306
                         boolean hasCswGetCapabilities = false;
                         Map<String, Boolean> pyCswStatus = new HashMap<>();
-                        String cswVersion = pycswVersion;
                         
-                        if ( (null == pycswVersion) || (pycswVersion.isEmpty()) ){
-                            cswVersion = pycswVersionDefault;
-                        }
-                        
-			String endpointTest = pycswEndpoint + "?service=CSW&request=GetCapabilities&version=" + cswVersion;
+			String endpointTest = pycswEndpoint + "?service=CSW&request=GetCapabilities&version=" + pycswVersion;
                         HttpGet httpGet = new HttpGet(endpointTest);
                         HttpClient httpclient = new DefaultHttpClient();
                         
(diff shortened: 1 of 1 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
```diff
@@ -130,7 +144,57 @@ public Response healthCheck() {
...
 			overallHealth = false;
 		}
-		
+                
+                try { // health check add for pycsw Jira cchs-306
+                        boolean hasCswGetCapabilities = false;
+                        Map<String, Boolean> pyCswStatus = new HashMap<>();
+                        String cswVersion = pycswVersion;
+                        
+                        if ( (null == pycswVersion) || (pycswVersion.isEmpty()) ){
```
cschroed-usgs: Could the assignment of the default value for pyCSWVersion be moved into the static initializer? There is a two-argument signature for `DynamicReadOnlyProperties#getProperty()` that allows specifying a default value.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
```diff
@@ -130,7 +144,57 @@ public Response healthCheck() {
...
+                                    return entity != null ? EntityUtils.toString(entity) : null;
+                                } else {
+                                throw new ClientProtocolException("Unexpected response status: " + status);
+                                }
+                            }
+                        }; // close anonymous inner class
+                    
+                        String responseBody = httpclient.execute(httpGet, responseHandler);
+                        
+                        if (responseBody.contains("Exception")){
```
cschroed-usgs: I used to take this same approach for detecting exceptions in web service responses in the past until it burned me bad. Another team ended up wasting a ton of time debugging because their data started containing "Exception", and I had written a false-positive-prone check like the one above. After several lashes with wet noodles, the best practice that emerged was to assume the response is correct, attempt to fully parse the response, and treat any parser exceptions as an indication of error; so taking more of the "It is better to beg for forgiveness than to ask for permission" approach.

There are legitimate reasons why the word "Exception" would occur in valid PyCSW GetCapabilities responses. We don't want to give off false alarms, so we should at least attempt a more rigorous check. All that said, I've never tried parsing a CSW GetCaps response in Java. Difficulty + deadlines can influence the rigor of our approach. Could you please try leveraging the JAXB PyCSW 2.0.2 bindings provided here?
http://www.ogcnetwork.net/jaxb4ogc
If that doesn't work, can you think of any halfway solutions that would be easier to implement, and less prone to reporting false health failures?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
smlarson-usgs: Thanks - didn't know that was readily available. I'll definitely take advantage of that feature.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
smlarson-usgs: I could add Exception Report ... as that's what's generated by the code to make it more specific then the generic exception. I believe the CCH code does this already. (I could avoid the 40 lashes ;)

smlarson-usgs:
Updates from the code review - Thanks Carl!

smlarson-usgs:
It appears my comment regarding the parsing of the response did not save previously. In essence, Megan put in the ticket that she wanted a simple health check. Jordan suggested that the external end point was used in the health check (see slack Jan 9 about 9:20 AM). The parsing of the response from the getCapabilities did not add much value since it would most likely not be functioning if the request threw an exception which was already handled like the rest of the code - failing the health check. Happy to do more if Megan feels its warranted. Thanks for your thoughtful comments on the review.

cschroed-usgs:
👍

PR #1538, opened 2017-01-12
Cchs 320

cschroed-usgs:
Resolving a maven dependency version conflict, upgrading to a new version of the proxy-utils package.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,7 +282,7 @@
 			<dependency>
 				<artifactId>proxy-utils</artifactId>
 				<groupId>gov.usgs.cida</groupId>
-				<version>1.0.12</version>
+				<version>1.0.13</version>
 				<type>jar</type>
 				<exclusions>
 					<exclusion>
--- a/pom.xml
+++ b/pom.xml
@@ -196,13 +196,8 @@
 			<dependency>
 				<groupId>org.apache.httpcomponents</groupId>
 				<artifactId>httpclient</artifactId>
-				<version>4.2.5</version>
+				<version>4.5.1</version>
 				<type>jar</type>
-			</dependency>
-			<dependency>
-				<groupId>org.apache.httpcomponents</groupId>
-				<artifactId>httpcore</artifactId>
-				<version>4.2.4</version>
 			</dependency>
 			<dependency>
 				<groupId>ch.qos.logback</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -275,9 +275,14 @@
 				<scope>provided</scope>
 			</dependency>
 			<dependency>
+				<groupId>org.slf4j</groupId>
```
cschroed-usgs: I added this because proxy-utils removed it from its pom. Without adding this, we were unintentionally leveraging jcl-over-slf4j as imported by the geoserver-manager artifact.

pom.xml
```diff
@@ -275,9 +275,14 @@
...
+				<artifactId>jcl-over-slf4j</artifactId>
+				<version>1.6.6</version>
```
duselmann: this 1.6.6 is a popular version, but I noticed that your slf4j.version is 1.7.13
just an FYI in case it was an oversight

duselmann:
(slf4j version fyi) ... This looks good otherwise.

pom.xml
cschroed-usgs: Ah, it was an oversight. Thanks for bringing that up. I had simply copied the jcl-over-slf4j pom entry that was removed from proxy-utils into the cch pom, without syncing version numbers.

pom.xml
```diff
@@ -196,15 +196,10 @@
 			<dependency>
 				<groupId>org.apache.httpcomponents</groupId>
 				<artifactId>httpclient</artifactId>
-				<version>4.2.5</version>
+				<version>4.5.1</version>
```
cschroed-usgs: proxy-utils requires httpclient 4.5.1. Upgrading this in the CCH parent pom makes the version required by the proxy-utils dependency the same as the version required by CCH's maven modules (coastal-hazards-commons, coastal-hazards-export).

pom.xml
```diff
@@ -196,15 +196,10 @@
...
 				<type>jar</type>
 			</dependency>
 			<dependency>
-				<groupId>org.apache.httpcomponents</groupId>
```
cschroed-usgs: This was no longer used directly -- it is only pulled in as a transitive dependency.

pom.xml
```diff
@@ -277,7 +272,7 @@
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>jcl-over-slf4j</artifactId>
-				<version>1.6.6</version>
+				<version>${slf4j.version}</version>
```
cschroed-usgs: Linking the slf4j versions so they don't get out of sync.

PR #1542, opened 2017-02-28
adding top three items to uber under bootstrap

mhines-usgs:
- adding the three main super items under uber
- modifying bootstrap to also add the aggregation_children records
- modifying bootstrap to start numbering later to leave room for the kids


mhines-usgs:
If you guys don't mind taking a gander at this for sanity check, and Zack, mayhaps you can try running locally to test? :/ thanks

coastal-hazards-liquibase/src/main/liquibase/bootstrap.xml
```diff
@@ -33,13 +33,19 @@
             <column header="show_children" name="show_children" type="BOOLEAN" />
             <column header="enabled" name="enabled" type="BOOLEAN" />
         </loadData>
+		<loadData encoding="UTF-8" file="bootstrap/aggregation_children.csv" quotchar="&quot;" separator="," tableName="aggregation_children">
+			<column header="aggregation_id" name="aggregation_id" type="STRING"/>
+			<column header="item_id" name="item_id" type="STRING"/>
+			<column header="list_index" name="list_index" type="NUMERIC"/>
+			<column header="id" name="id" type="NUMERIC"/>
```
jiwalker-usgs: I think this either needs to be removed (so db defaults it) or it should be added manually in data below.

mhines-usgs: ok, i'll remove it. thanks for pointing that out. and tabs, woops. fixing.

PR #1545, opened 2017-03-07
Cchs 189

kmschoep-usgs:


coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -336,6 +337,12 @@ CCH.Objects.Publish.UI = function () {
 			} else if ($titleMediumTextArea.val().length > CCH.CONFIG.limits.summary.medium.title) {
 				errors.push('Medium title was longer than ' + CCH.CONFIG.limits.summary.medium.title + ' characters');
 			}
+			
+			if (me.isBlank($titleLegendTextArea)) {
```
zmoore-usgs: Should this isBlank call be done on `$titleLegendTextArea.val()`?

-------------------------------------------------------------------------

Also I'm not sure if this matter or not, and I might also be completely wrong in how this works since I can't see all of the code in the github view, but should the isBlank call be done on the trimmed version of the text in the text area? As it is currently it seems like no errors would show up if I provided a title of `\s\s\s\s` (where `\s` is a space) but then when the title is saved on line 427 it is trimmed down to an empty string. 

This probably applies to all of the existing form fields as well, so not something specifically wrong with your code but more just a comment on the pattern overall.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -336,6 +337,12 @@ CCH.Objects.Publish.UI = function () {
...
+				errors.push('Legend title not provided');
+			} else if ($titleLegendTextArea.val().length > CCH.CONFIG.limits.summary.legend.title) {
```
zmoore-usgs: Same thought as above. Should this be done on the trimmed value?

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -111,7 +111,7 @@ CCH.Objects.Widget.Legend = function (args) {
 			$theadUOM = $('<td />'),
 			bins = sld.bins,
 			uom = sld.units || '',
-			title = (args.item && args.item.summary && args.item.summary && args.item.summary.medium) ? args.item.summary.medium.title : sld.title || ''
+			title = (args.item && args.item.summary && args.item.summary && args.item.summary.legend) ? args.item.summary.legend.title : sld.title || ''
```
zmoore-usgs: Should `args.item.summary`  be repeated or is one of those supposed to be something else? Looks like it was repeated originally so maybe the duplicate can just be removed?

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
```diff
@@ -162,7 +162,7 @@ CCH.Objects.Widget.Legend = function (args) {
 			$rangeTd,
 			bins = sld.bins,
 			uom = sld.units || '',
-			title = (args.item && args.item.summary && args.item.summary && args.item.summary.medium) ? args.item.summary.medium.title : sld.title || '',
+			title = (args.item && args.item.summary && args.item.summary && args.item.summary.legend) ? args.item.summary.legend.title : sld.title || '',
```
zmoore-usgs: Same as above

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/summary/Legend.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.persistence.Embeddable;
+
+/**
+ *
+ * @author Kathryn Schoephoester <kmschoep@usgs.gov>
+ */
+@Embeddable
+public class Legend implements Serializable {
+
+    public static final int TITLE_MAX_LENGTH = 40;
```
mhines-usgs: I would vote for this to be longer.
I know the idea is to keep it short, but we have had a lot of headaches dealing with running out of room. Providing the location in my opinion is our job, but they should be in charge of keeping it short. Granted, this is my opinion.

coastal-hazards-liquibase/src/main/liquibase/addSummaryLegendTitle.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd" logicalFilePath="liquibaseChangeSet/addPublicationType.xml">
+    <changeSet author="kmschoep" id="add_legend_title">
+        <addColumn tableName="summary">
+            <column name="legend_title" type="VARCHAR(40)"/>
```
jiwalker-usgs: I should have mentioned this yesterday, but my pattern of using `VARCHAR(n)` is apparently not necessary in postgres.  Making it a `TEXT` field does not have any real trade offs in size/performance.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/summary/Legend.java
kmschoep-usgs: Yeah, 40 was my first guess.  How long do you think is reasonable?

jiwalker-usgs: I think we can remove the limit altogether if we go with text field.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/Legend.js
jiwalker-usgs: yeah, that is weird that it was this way.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/summary/Legend.java
mhines-usgs: I would second turning it into Text and removing all limitations! Then we never have to cut another release related to changing length needs (for this field, at least!)

kmschoep-usgs: What does "turning it into Text" mean, just the UI?  a postgres data type?

mhines-usgs: yeah it's a postgres data type without any character limit

kmschoep-usgs: ok

mhines-usgs: we have that on service.service_endpoint if you're curious what it looks like.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
kmschoep-usgs: You have a point!  I'll do some testing to see if I can harden that form a bit.

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
```diff
@@ -261,7 +268,7 @@
 							</div>
 							<div id="form-publish-info-item-description-tiny" class="row row-description">
 								<div class="form-group">
-									<label for="form-publish-item-description-tiny"><span class="emphasis-aggregation"><i class="fa fa-asterisk"></i></span>Description (Tiny) / Legend Title and Tweet Text</label>
+									<label for="form-publish-item-description-tiny"><span class="emphasis-aggregation"><i class="fa fa-asterisk"></i></span>Description (Tiny) / Tweet Text</label>
```
mhines-usgs: :D

PR #1546, opened 2017-03-07
CCHS-289: Modified click to identify behavior to not...

zmoore-usgs:
... average the clicked values and instead show the top 3 features.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
@@ -212,7 +212,7 @@ return {
 			units = sld.units,
 			layerId = args.layerId,
 			item = CCH.items.getById({id: layerId}),
-			title = item.summary.medium.title,
+			title = item.summary.legend.title,
 			features = args.features,
 			attr = overrideAttributeName(features, item.attr),
 			displayPoints = new Array(),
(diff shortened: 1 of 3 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -432,48 +453,50 @@ return {
...
+					
+					//If we have added a 3rd value stop trying to add more
+					if(displayPoints.length === 3){
+						break;
 					}
 				}
 			}
 		}
+		
+		if (["TIDERISK", "SLOPERISK", "ERRRISK", "SLRISK", "GEOM", "WAVERISK", "CVIRISK", "AE"].indexOf(item.attr.toUpperCase()) !== -1) {
```
jiwalker-usgs: I'm not sure what makes these special, but if we can it would be good to avoid hardcoding attributes in the UI.  I know this is just carried over, but thinking about how this might be pulled out would be useful.

zmoore-usgs: Yeah, I'm not sure why these specifically are pulled out. Are they all raster attributes? 

I tried taking this completely out earlier and it resulted in the raster data popups being off by 1 "bin". I.E: Clicking on a dark blue area on the map creates a legend representing a light blue area, clicking light blue results in dark green, etc.

jiwalker-usgs: I think these are the "binned" attributes.  AE is one of the raster attributes, but I don't exactly see why this would be a thing.

PR #1548, opened 2017-03-09
CCHS-329: Allow users to upload vector shapefiles and their XML metadata

zmoore-usgs:


coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/vector.jsp
```diff
@@ -0,0 +1,103 @@
...
+<%@page import="gov.usgs.cida.coastalhazards.model.Service"%>
+<%@page import="gov.usgs.cida.coastalhazards.model.Item"%>
+<%@page contentType="text/html" pageEncoding="UTF-8"%>
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
+<%@page import="org.apache.commons.lang.StringUtils"%>
+<%@page import="gov.usgs.cida.config.DynamicReadOnlyProperties"%>
+<%@page import="java.util.Map" %>
+
+<%!
+	protected DynamicReadOnlyProperties props = new DynamicReadOnlyProperties();
```
thongsav-usgs: Might be a good idea to extract this out to a Java class/singleton. Not much logic here, so no biggie, but more room for runtime failures  here without more graceful error handling/output.

coastal-hazards-portal/src/main/webapp/js/application/publish/vector.js
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * 
+	 * @param {String} layerUrl
+	 * @returns {String} the layer id
+	 */
+	var getLayerIdFromUrl = function(layerUrl){
+		return layerUrl.from(layerUrl.lastIndexOf('/') + 1);
+	};
+	$("#upload-btn").click(function(e){
+		var $result = $('#result');
```
thongsav-usgs: More of an FYI, but I personally don't like using html ids as you are limited to one instance per page. Using html classes allows more OO (multi-instance) logic.

zmoore-usgs:
@thongsav-usgs The jsp is pretty much just a direct copy-paste of the raster one and is just a placeholder to test functionality for now. I agree with both of your comments, but I think I'll wait and address them in the next ticket I'm starting on today since the vector.jsp and raster.jsp pages will be completely going away with that.

PR #1550, opened 2017-03-15
CCHS-322: Create new field for data downloads

zmoore-usgs:


mhines-usgs:
are you comfortable merging these, @zmoore-usgs ?

zmoore-usgs:
Yep, sorry was doing training and didn't see this was approved.

PR #1551, opened 2017-03-15
Layer Select Pop Up Less

mwernimont:
1. Changed the layer-identify-close.svg to black to better match the mockup and contrast benefits.

2. Removed the .css('color', 'white')); from LayerIdentifyControl.js to have the browser use the inheirt black text color and remove unnecessary code after the color change.

3.  Made the pop up match the mockup closer.

4. Removed that image from 2016, guess I haven't worked on CCH in awhile...

@zmoore-usgs look alright?


zmoore-usgs:
@mwarren-USGS Code changes look good to me. Tested locally and everything seems to work well, just one bit of feedback: 

If you click on the map when you are in the bucket view and have multiple layers selected the rows added from new layers all start white so you can end up with several white rows sequentially in the table without any sort of visual break between them. This doesn't bother me since the layer name column value changes and that provides a break, but just something to consider. See screenshot below.

color

mwernimont:
@zmoore-usgs Yeah I talked to @mhines-usgs about that, she thinks it will be alright for now, but if PO's get the itch for that to change I have a plan. Also just so Meredith doesn't get dragged to everything we work on together its @mwernimont. lol. :)

mhines-usgs:
I think it's oK for now even if the striping is weird; the cooperators provided feedback which might require further tinkering with the click to ID logic. I'm trying to have them discuss and give us firmer details for revisions and we can include improved striping at that point if needed. Thanks guys, I'd be ok merging this as is for now.

mwernimont:
Up to you boss @mhines-usgs

mhines-usgs:
:shipit: :airplane: :rocket:

PR #1557, opened 2017-03-28
mediation page makeover

mwernimont:
@zmoore-usgs or @thongsav-usgs  hows it look to you?
Tested it on Chrome, FireFox, and Safari, and an ipad, the layout looked the same on all of them. Hopefully, the shifting of content didn't break any functionality.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
+++ b/coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/publish/item/index.jsp
@@ -18,622 +18,674 @@
 <%@page import="java.util.Map" %>
 
 <%!
-	protected DynamicReadOnlyProperties props = new DynamicReadOnlyProperties();
+    protected DynamicReadOnlyProperties props = new DynamicReadOnlyProperties();
 
-	{
-		try {
-			File propsFile = new File(getClass().getClassLoader().getResource("application.properties").toURI());
-			props = new DynamicReadOnlyProperties(propsFile);
-			props = props.addJNDIContexts(new String[0]);
-		} catch (Exception e) {
-			System.out.println("Could not find JNDI - Application will probably not function correctly");
-		}
-	}
+    {
+        try {
+            File propsFile = new File(getClass().getClassLoader().getResource("application.properties").toURI());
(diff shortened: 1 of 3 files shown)


zmoore-usgs:
Code changes look good to me. I'll pull down the changes and test the functionality locally

mhines-usgs:
Ah I didn't think about the potential for expanding resources darn.. @zmoore-usgs, good call.

I guess moving the metadata section to the bottom left is better since the keywords would probably suffer from the same problem.. Then because it'd be left-heavy, move titles/descriptions above the bounding box/keywords panel? What do you guys think of that?

mwernimont:
@mhines-usgs So move metadata underneath titles/descriptions?

mhines-usgs:
I would say switch columns and locations for titles/descriptions and metadata. That way metadata is at the bottom of the left column and bounding boxes and keywords are at the bottom of the right column. Both of those have a potentially infinite length?

zmoore-usgs:
@mhines-usgs @mwernimont Would there be a way to make it ordered in the way that each step needs to be completed? II.E: I think the order of steps is roughly:

1). Create item
2). Create Layer
3). Populate Services from Layer
4). Select Item Type & Attribute
5). Populate Metadata
5). Titles/Descriptions, Thumbnail, and Resources

So maybe something like:

Item ID     | Tumbnail 
Services    | Titles / Descriptions
Item Type | Resources
Metadata  |

Thoughts? It's okay if you guys disagree, just a thought. Then the flow is go down the left side and then go down the right side. Maybe even number each box so that it is clear what the order needs to be?

mwernimont:
Talked to @mhines-usgs and decided to just have the possible ever expanding options (Resources Keywords) on the bottom of the either column. 

Bueno?

mhines-usgs:
Yeah, that's good. Can you also edit out the validation part for download file name that @zmoore-usgs mentioned? I forgot about that. We don't need that anymore too. Thank you!

mwernimont:
@zmoore-usgs that was the correct code block to remove right? Hopefully our IDE's line up the code the same.

zmoore-usgs:
@mwernimont Yep, looks good to me. I'll pull down the changes and test it out again real quick but then I think this should be good to go.

PR #1558, opened 2017-03-28
Couple of minor clean up fixes, but main change is the timestamp change

thongsav-usgs:


mhines-usgs:
OK to merge then @thongsav-usgs ?

PR #1560, opened 2017-03-30
CCHS-344 update for download link for pcoi items

mhines-usgs:
tested with two items on dev and it worked ok there.


mhines-usgs:
Discussed on the JIRA ticket that we are kind of using liquibase incorrectly for this need. I'll try to come up with an R solution instead...

PR #1564, opened 2017-03-30
CCHS-191: Fixed a PostGresSQL connection leak caused by the recursive…

zmoore-usgs:
… item filtering for historical legends.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Shorelines.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Shorelines.java
@@ -46,9 +46,8 @@ public final class Shorelines {
 		}
 
 		public void finalize(Item item) {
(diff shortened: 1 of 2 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Shorelines.java
```diff
@@ -48,13 +48,17 @@ public ShorelineConfig(String jspPath, String units, String style, int strokeWid
 		public void finalize(Item item) {
 			try (DataDomainManager manager = new DataDomainManager()) {
 				DataDomain domain = manager.getDomainForItem(item);
+				ItemManager items = new ItemManager();
```
jiwalker-usgs: You can put this in the try-with-resources.  You shouldn't need the `close()` statements if you do that (they are handled by an implicit finally).

PR #1565, opened 2017-03-31
Maintain zoom level in session so that the shared views remains appear

thongsav-usgs:
as intended

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/Session.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/Session.js
@@ -22,6 +22,7 @@ CCH.Objects.Session = function (args) {
 		baselayer: '',
 		scale: 55467893.20,
 		bbox: [-138.48502349853044, 7.986650381678925, -55.86783599853124, 56.56752394039768],
+		zoomLevel: 3,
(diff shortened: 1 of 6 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/Session.js
```diff
@@ -22,6 +22,7 @@ CCH.Objects.Session = function (args) {
 		baselayer: '',
 		scale: 55467893.20,
 		bbox: [-138.48502349853044, 7.986650381678925, -55.86783599853124, 56.56752394039768],
+		zoomLevel: 3,
```
mhines-usgs: is this a default for old sessions?

thongsav-usgs: No, it's the default for when the app loads for the first time. For the old shared views it's going to do the old behavior of being one zoom level too far. I can probably throw in a hack and always zoom in one level if zoom level isn't saved, but there are some cases where that will zoom in too far.

mhines-usgs: That's ok, just was wondering. If the old sessions behave as they were, that is fine. Thanks!

PR #1567, opened 2017-04-03
Include tiny text in bucket item share via tweet

thongsav-usgs:


coastal-hazards-portal/src/main/webapp/js/cch/objects/front/UI.js
```diff
@@ -165,7 +165,7 @@ CCH.Objects.Front.UI = function (args) {
 					hashtags: 'USGS_CCH',
 					lang: 'en',
 					size: 'large',
-					text: 'Check out my CCH View!',
+					text: shareTitle || 'Check out my CCH View!',
```
mhines-usgs: I think the intent is that the Tweet text is the only text, drop the "Check out my CCH View part altogether", that would make it too long to be a Tweet

mhines-usgs: image

zmoore-usgs: @mhines-usgs I think this line is saying that the text will be the shareTitle if it exists, and otherwise will use "Check out my CCH View!" as a default value if the shareTitle (tweet text) is not defined.

thongsav-usgs: That's what that does, the code defaults to "Check out my CCH View" for the share view function.

mhines-usgs: ah, gotcha, then :+1: -- sorry about the confusion, I wasn't sure if that was appending it or what. Thanks!

PR #1568, opened 2017-04-03
CCHS-188 / CCHS-191: Fixed some issues where the map and...

zmoore-usgs:
... click-to-identify colors were not matching the legend colors.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDGenerator.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDGenerator.java
@@ -72,7 +72,7 @@ public class SLDGenerator {
 		}
 	}
 	
-	public static SLDGenerator getGenerator(Item item, Integer ribbon) {
+	public static SLDGenerator getGenerator(Item item, Item selectedItem, Integer ribbon) {
 		SLDGenerator generator = null;
         
         Item.Type itemDotType = item.getType();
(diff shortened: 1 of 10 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Shorelines.java
```diff
@@ -45,8 +45,12 @@ public ShorelineConfig(String jspPath, String units, String style, int strokeWid
 			wrapped = null;
 		}
 
-		public void finalize(Item item) {
+		public void finalize(Item item, Item selectedItem) {
```
thongsav-usgs: Did the signature have to change for this? Doesn't look like selectedItem is used anywhere except to overwrite item. Also seems like item wasn't used before being overwritten. Couldn't it be up to callers of this method to pass in the correct item?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDGenerator.java
```diff
@@ -114,12 +114,12 @@ public static boolean isValidAttr(SLDConfig config, String attr) {
...
 	 * @param config 
 	 */
-	protected SLDGenerator(Item item, Integer ribbon, SLDConfig config) {
+	protected SLDGenerator(Item item, Item selectedItem, Integer ribbon, SLDConfig config) {
 		this.item = item;
 		this.ribbon = ribbon;
         if (config instanceof ShorelineConfig) {
             ShorelineConfig slc = (ShorelineConfig)config;
-            slc.finalize(item);
+            slc.finalize(item, selectedItem);
```
thongsav-usgs: Seems at this point you could just pass in the selectedItem instead of both.

jiwalker-usgs:
Is there a way to test the new behavior to show how it is supposed to work? This looks good, but a couple of the item vs. selectedItem bits are somewhat difficult to follow.

PR #1572, opened 2017-04-10
Geoserver Module now just collects needed config, no longer builds deployable war

thongsav-usgs:


coastal-hazards-geoserver/pom.xml
```diff
@@ -36,56 +29,47 @@
...
-                        src/main/webapp/data/workspaces/upload,
-                        src/main/webapp/data/workspaces/draw
-                    </packagingExcludes>
-                    <failOnMissingWebXml>false</failOnMissingWebXml>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
+	<build>
+		<finalName>geoserver</finalName>
```
jiwalker-usgs: I suggest a more descriptive finalName.

coastal-hazards-geoserver/src/assembly/README.md
```diff
@@ -0,0 +1,10 @@
+This zip file contains custom configuration files for Geoserver ${gs.version}.
```
jiwalker-usgs: Do these tokens get replaced?

coastal-hazards-geoserver/src/assembly/geoserver-config-assembly.xml
```diff
@@ -0,0 +1,44 @@
...
+			<directory>${project.basedir}/src/main/webapp/data</directory>
+			<outputDirectory>data</outputDirectory>
+		</fileSet>
+		<fileSet>
+			<directory>${project.basedir}/src/assembly</directory>
+			<outputDirectory>/</outputDirectory>
+			<includes>
+				<include>README.md</include>
+			</includes>
+			<filtered>true</filtered>
```
jiwalker-usgs: Ignore above, I see the filter happens here.

coastal-hazards-geoserver/src/assembly/geoserver-config-assembly.xml
```diff
@@ -0,0 +1,44 @@
...
+		</fileSet>
+	</fileSets>
+	<dependencySets>
+		<dependencySet>
+			<outputDirectory>/lib</outputDirectory>
+			<useProjectArtifact>true</useProjectArtifact>
+			<unpack>false</unpack>
+			<scope>runtime</scope>
+			<includes> <!-- only include things that are not provided by tomcat/geoserver -->
+				<include>gov.usgs.cida.coastalhazards:coastal-hazards-commons</include>
```
jiwalker-usgs: I feel like maintaining this list is going to be a nightmare.  Did you think about packaging a jar with dependencies?  I think separating the functional components from the data/configuration would be useful.  Putting them in the same zip seems to kind of break the purpose of using maven in my mind.

coastal-hazards-geoserver/pom.xml
```diff
@@ -7,10 +7,10 @@
         <version>1.1.53-SNAPSHOT</version>
     </parent>
 
-    <artifactId>coastal-hazards-geoserver</artifactId>
-    <packaging>war</packaging>
-
-    <name>Coastal Hazards Geoserver</name>
+	<artifactId>coastal-hazards-geoserver</artifactId>
+	<packaging>pom</packaging>
```
thongsav-usgs: Yeah, I wasn't sure how to make that happen. Maven said zip wasn't valid. I'll look more into it, though.

coastal-hazards-geoserver/pom.xml
thongsav-usgs: :+1:

coastal-hazards-geoserver/src/assembly/geoserver-config-assembly.xml
thongsav-usgs: Unfortunately it's the same issue with war/jar packaging in that the full list of dependencies is included. It was 160 ish jars versus the 10 or so listed here in the war previously, and some libs were packaged multiple times (different versions). 

It definitely will be a pain keeping it up to date, but that may be a good thing given it'll force us to revisit and review the list.

PR #1574, opened 2017-04-13
New Item Order of Operations

mwernimont:
@zmoore-usgs 

So I reformatted the file in netbeans to get a cleaner look in my IDE and didn't realize github would consider it a whole new file, I'll try to describe what I did to help with that mess.

First how the new item page should work.
1.) Select New Item > Item : Only the Services Panel should be ungrayed.

2.) Once you fill in the Populate from Layer ex FauACWJh; and hit publish, the Item Type and Entire Feature panel unlocks. Should also unlock when you do create vector layer and hit Populate Services From Layer after you find the shapelayer you want to use.

3.) Selecting the Item type unlocks the Get Attributes Button.

4.) Clicking the Get Attributes Button unlocks both the Attributes input and the Type above (raster) or choose below (other) select.

5.) Either typing directly in the Attributes input (3 characters minimum) or selecting an option from the Type above (raster) or choose below (other) select, will unlock the rest of the options, which are Titles, Resources, and Metadata.

Functions created

line 750 unlockItemTypeFeatures

line 758 unlockTitlesReourcesMetadata

line 764 lockTitlesReourcesMetadata

2175 a .on input for $attributeSelect

Any ideas of improvement for the functions or how to fix this reformatted check in all the friggin lines annoyance that I did would be appreciated.


mhines-usgs:
One thought, could you apply the same validation when the user uses the create raster layer button that you extended to the create vector layer button action? (If you didn't already?)

mwernimont:
@mhines-usgs Good thinking

mwernimont:
Ugh, reverting the branch. Dumb spaces

PR #1577, opened 2017-04-13
New Item Order

mwernimont:
How the new item page should work.
1.) Select New Item > Item : Only the Services Panel should be ungrayed.

2.) Once you fill in the Populate from Layer ex FauACWJh; and hit publish, the Item Type and Entire Feature panel unlocks. Should also unlock when you do create vector layer and hit Populate Services From Layer after you find the shapelayer you want to use.

3.) Selecting the Item type unlocks the Get Attributes Button.

4.) Clicking the Get Attributes Button unlocks both the Attributes input and the Type above (raster) or choose below (other) select.

5.) Either typing directly in the Attributes input (3 characters minimum) or selecting an option from the Type above (raster) or choose below (other) select, will unlock the rest of the options, which are Titles, Resources, and Metadata.

@zmoore-usgs Looks like I got it fixed

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
@@ -742,15 +742,33 @@ CCH.Objects.Publish.UI = function () {
 	me.updateSelectChange = function () {
 		if ($attributeSelectHelper.val() !== '') {
 			$attributeSelect.val($attributeSelectHelper.val());
+			me.unlockTitlesResourcesMetadata();
 		}
 	};
         
-        me.itemTypeCheck = function(){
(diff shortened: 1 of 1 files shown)


mwernimont:
@zmoore-usgs Switched the functionality to unlock the attribute input when you hit publish on any of the 3 options.

The Titles, Resources, and Metadata buttons unlock when you load in an item now.

I do have a question though, when you load in an extreme storm services is all locked? Expected functionality, cause I don't remember doing anything for that to happen. It has no entries in the services so It seems like that's what it's supposed to do, but want to make sure.

zmoore-usgs:
@mwernimont I believe that is the intended behavior for storms because the storms are Template items which really just hold references to the actual data items that are stored underneath them in the tree, so I don't think the Template itself would have services. 

I'd defer to @mhines-usgs for confirmation of that though. I'm pulling the updates locally and testing now.

PR #1581, opened 2017-04-20
Character min limit changed

mwernimont:
Was informed there is shorter Attribute names than 3 characters


mwernimont:
@zmoore-usgs Simple change for Attribute character check to unlock Titles, Resources, and Metadata.

If you have a more elegant solution to this as a whole, let me know, I'm just acting like I know programming.

PR #1583, opened 2017-04-21
CCHS-377: Updated only-visible-children filtering to...

zmoore-usgs:
... be toggleable via the API call. Defaults to false.


zmoore-usgs:
This is now incorporated into the general performance profiling PR.

PR #1585, opened 2017-05-05
CCHS-384: Updated data domains to only build from visible children. …

zmoore-usgs:
…Updated data domain code to delete all domains when the tree is updated (may want to change this method later).

This PR is a bit of a WIP but I wanted to get some feedback on it so far. I included a comment on one section that I'm not totally sure about.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/DataDomainManager.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/DataDomainManager.java
@@ -26,6 +26,7 @@ public class DataDomainManager implements AutoCloseable {
     
     private static final String HQL_SELECT_BY_ID = "select d from DataDomain d where d.itemId = :id or d.sessionId = :id";
     private static final String HQL_DELETE_ALL= "DELETE FROM DataDomain";
+    private static final String HQL_DELETE_BY_ID = "DELETE from DataDomain where itemId = :id or sessionId = :id";
     private static final Map<String, Lock> locks = Collections.synchronizedMap(new HashMap<String, Lock>());
     
     private EntityManager em;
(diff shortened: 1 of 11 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/ItemManager.java
```diff
@@ -165,6 +165,8 @@ private synchronized String mergeItem(Item item) {
 			item.setChildren(replaceList);
 		}
 		em.merge(item);
+		DataDomainManager dm = new DataDomainManager();
+		dm.deleteAll();
```
zmoore-usgs: Currently I'm deleting all values from DataDomains whenever the tree is updated. This could probably be restricted to at least only happening when the item has a type of 'historic' (though I'm not sure how magic-gnome-like that is) but aside from that every Item upstream of the two modified items would need to have its domain cleared so I'm not sure if it would be worth the processing cost to traverse and identify the upstream domains versus just regenerating them all. Domains for Items are currently generated when the `getDomainForItem` function is called and the domain for the provided item doesn't not exist yet.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/DataDomainManager.java
```diff
@@ -25,6 +25,7 @@
     private static final Logger log = Logger.getLogger(DataDomainManager.class);
     
     private static final String HQL_SELECT_BY_ID = "select d from DataDomain d where d.itemId = :id or d.sessionId = :id";
+    private static final String HQL_DELETE_ALL= "DELETE FROM DataDomain";
```
zmoore-usgs: I was thinking about that but I'm not sure if savings in only re-calculating some domains would outweigh with the added overhead of hunting down all of the domains that need to be deleted?

Every Item upstream of our updated item's original position that had it visible would need to have its domain deleted, as well as every upstream Item that has our item visible from its new position. This would need to take into account that not every Item upstream of our updated item may have it visible (i.e: Aggregation C could have it visible and Aggregation B could have Aggregation C visible while Aggregation A does not have Aggregation B visible meaning B and C would need to update but A does not). 

Going down the tree is pretty easy to do but it doesn't seem like the items have a link to their parent? If that's the case then I think traversing back up would be more difficult and would require some searching and that could include some unnecessary item loads.

jiwalker-usgs: I'm just wondering what costs more, doing a more selective delete, or repopulating all data domains?

zmoore-usgs: I'm thinking the selective delete will end-up costing more because of the tree not being doubly-linked but I'll mess around with it a bit and compare

zmoore-usgs: I've just about finished a recursive function to delete the upstream domains of an item but it actually looks like there is a SQL stored procedure to get all of the upstream ancestors of an item. 

I could create a similar one that limits the scope to ancestors that have the item visible in their domain but considering that there are only two stored procedures that I can find I wasn't sure if we are trying to get away from using them?

zmoore-usgs:
@jiwalker-usgs I updated this to use a stored procedure in order to find and delete only the relevant data domains.

PR #1586, opened 2017-05-09
CCHS-384: Updated data domains to be regenerated instead of just deleted.

zmoore-usgs:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/DataDomainResource.java
```diff
@@ -49,4 +53,22 @@ public Response getDataDomain(@PathParam("id") String id, @Context Request reque
...
+    @GET
+    @RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
+    @Path("/regenall")
+    public Response regenerateAllDataDomains() {
+        Response response = null;
+        try (ItemManager itemManager = new ItemManager(); DataDomainManager domainManager = new DataDomainManager()) {
+	    List<Item> rootItems = itemManager.loadRootItems();
+	    
+	    if(rootItems.size() == 1){
+		List<String> generatedIds = domainManager.regenerateAllDomains(rootItems.get(0));
```
thongsav-usgs: How slow is this? I was wondering if it's maybe a better idea to always do this when the application launches as well.

zmoore-usgs: I haven't tried it with as many items as DEV yet. Locally it takes about 15 seconds with 6 historical items. 

I'm not sure if it would need to do it every time the application launches because the only time these would be invalidated would be if the algorithm to generate them changes or if the tree is modified (which itself will cause the regeneration). However that would negate the manual step of running it the first time this change is deployed to a tier.

PR #1591, opened 2017-05-12
CCHS-184 add new SLD for polygons

kmschoep-usgs:

PR #1592, opened 2017-05-15
CCHS-184: Create new SLD for polygon data

zmoore-usgs:


zmoore-usgs:
Currently working on addressing feedback from: https://github.com/USGS-CIDA/coastal-hazards/pull/1591

coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/SLD/evic.jsp
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jiwalker-usgs: Is this just a copy of bins_polygon.jsp?  There isn't a problem if EVIC uses polygon, it just needs to link to it in the SLDConfig.

zmoore-usgs: It has some tweaks. bins_polygon uses the thresholds directly to calculate things and seems to have a lot more going on. This uses the defined upper and lower bounds that are calculated in the SLDConfig

jiwalker-usgs: So is bins_polygon used anywhere?

zmoore-usgs: Doesn't look like it, so I could just edit that I suppose.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/Evic.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static gov.usgs.cida.coastalhazards.Attributes.*;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ *
+ * @author Kathryn Schoephoester <kmschoep@usgs.gov>
```
isuftin: Copypasting?

mhines-usgs: no, Kathy actually did this...

isuftin: Right but why is Zack checking it in?

isuftin: Ah I see now. @kmschoep-usgs needs a recognizable avatar 🥇 D

mhines-usgs: get back to work ;) :shoo:

PR #1596, opened 2017-05-19
Adjusting stroke width for line to 50, after getting some positive fe…

mhines-usgs:
…edback from Emily regarding her QA testing where it was set to 5.


mhines-usgs:
Pushed to 65
https://trello.com/c/kcb3oHJe

PR #1598, opened 2017-05-22
CCHS-397: Fixed an issue where the health check would report healthy …

zmoore-usgs:
…but return unhealthy for the CSW getCapabilities check.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
@@ -172,7 +172,7 @@ public class HealthResource {
                         String resp = httpclient.execute(httpGet, responseHandler); 
 			hasCswGetCapabilities = resp != null;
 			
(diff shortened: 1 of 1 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/HealthResource.java
```diff
@@ -169,9 +169,10 @@ public String handleResponse(
                             }
                         }; // close anonymous inner class
                     
-                        httpclient.execute(httpGet, responseHandler);                        
-                        pyCswStatus.put("getCapabilities", true);
-                      			
+			String resp = httpclient.execute(httpGet, responseHandler); 
+			hasCswGetCapabilities = resp != null;
```
jiwalker-usgs: You might want to check the response for issues, but this will fulfill the happy path.

zmoore-usgs: The response handler for this request does some error checking, but yeah probably would be a good idea to check the context of the returned xml a bit more closely.

PR #1599, opened 2017-05-22
CCHS-379: Fixed a potential connection leak if an item's domain …

zmoore-usgs:
…generation failed.


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/ItemManager.java
```diff
@@ -153,33 +153,38 @@ public synchronized boolean mergeAll(List<Item> items) throws CycleIntroductionE
...
-		
-		//Clear the upstream data domains if this item has a domain
-		if(dm.load(id) != null){
-		    List<Item> domainsToDelete = findVisibleAncestors(item);
-		    
-		    for(Item toDelete : domainsToDelete){
-			dm.regenerateDomainForItem(toDelete);
 		    }
+
+		    dm.close();
```
jiwalker-usgs: It should close on its own, no need to close it.

PR #1601, opened 2017-06-16
CCHS-336: Investigate upgrade pycsw version before spindling new stacks

zmoore-usgs:
Adds support for pycsw 2.x and keeps support for pycsw 1.x to smooth the transition. 1.x code should be removed once servers have been upgraded.


mhines-usgs:
I can grab a bunch of data from prod's geoserver to test with

PR #1606, opened 2017-07-21
CCHS-388: UI changes for creating/managing aliases

zmoore-usgs:


zmoore-usgs:
Forgot @kmschoep-usgs is out next week

PR #1609, opened 2017-09-07
CCHS-389: UI updates for any location that would use an alias

zmoore-usgs:


mhines-usgs:
I can't offer a lot here, but I approve.
One thing to remember about merging this is https://usgs-cida.slack.com/archives/C03M3F41K/p1501514745068245
recently, we did dev machine restarts and it's been throwing errors again and filling up the logs. Shirley just noticed it and I said we had it shut off. I just stopped that tomcat again but that means when you merge it won't auto deploy.
https://internal.cida.usgs.gov/jira/browse/CCH-74
I think the issue was that we didn't have it secured based on that slack thread.

zmoore-usgs:
Gotcha, I'll turn the tomcat back on manually to do the deploy to DEV. I can go ahead and try to shut off those old jobs and see what happens.

PR #1611, opened 2017-09-22
CCHS (MLR-160): First round of additional logging for vector layer uploads

zmoore-usgs:


coastal-hazards-portal/src/main/java/gov/usgs/cida/utilities/StringPrecondition.java
```diff
@@ -16,7 +16,7 @@ public static boolean willFit(String string, int length) {
     
     public static void checkStringArgument(String string, int length) {
         if (!willFit(string, length)) {
-            throw new IllegalArgumentException("Max length is: " + length);
+            throw new IllegalArgumentException("Max length is: " + length + " | Provided string (" + string.length() + "): " + string);
```
mhines-usgs: nice, yes this should help :D

PR #1614, opened 2017-11-28
CCHS-202: Let them delete items!

zmoore-usgs:
@mbucknell If I remember right CCH got mapped (or presumably mapped) to your branch for sometime in the future so I figured I add you on here as well in case you were interested in taking a look.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
@@ -216,22 +216,28 @@ public class ItemResource {
 	@RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
 	@DELETE
 	@Path("/{id}")
-	public Response deleteItem(@Context HttpServletRequest request, @PathParam("id") String id) {
+	public Response deleteItem(@Context HttpServletRequest request, @PathParam("id") String id, @QueryParam("deleteChildren") boolean deleteChildren) {
 		Response response = null;
+		
 		try (ItemManager itemManager = new ItemManager()) {
-			if (itemManager.delete(id)) {
-				response = Response.ok().build();
-			} else {
-				throw new Error();
-			}
-			try (StatusManager statusMan = new StatusManager(); ThumbnailManager thumbMan = new ThumbnailManager()) {
-				Status status = new Status();
-				status.setStatusName(Status.StatusName.ITEM_UPDATE);
-				statusMan.save(status);
+			if(itemManager.isOrphan(id)){
+				if (itemManager.delete(id, deleteChildren)) {
+					response = Response.ok().build();
+				} else {
+					throw new Error();
+				}
+				try (StatusManager statusMan = new StatusManager(); ThumbnailManager thumbMan = new ThumbnailManager()) {
+					Status status = new Status();
+					status.setStatusName(Status.StatusName.ITEM_UPDATE);
+					statusMan.save(status);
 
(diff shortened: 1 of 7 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -216,22 +216,28 @@ public Response updateItem(@Context HttpServletRequest request, @PathParam("id")
...
+					Status status = new Status();
+					status.setStatusName(Status.StatusName.ITEM_UPDATE);
+					statusMan.save(status);
 
-				thumbMan.updateDirtyBits(id);
-			}
+					thumbMan.updateDirtyBits(id);
+				}
+			} else {
+				response = Response.status(400).build();
```
mbucknell: I'm wondering why the status is 400 here rather than 404?

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -54,6 +54,10 @@ CCH.Objects.Publish.UI = function () {
 		$alertModalTitle = $alertModal.find('.modal-title'),
 		$alertModalBody = $alertModal.find('.modal-body'),
 		$alertModalFooter = $alertModal.find('.modal-footer'),
+		$deleteErrorModal = $('#delete-modal'),
+		$deleteErrorModalTitle = $deleteErrorModal.find('.modal-title'),
+		$deleteErrorModalBody = $deleteErrorModal.find('.modal-body'),
+		$deleteErrorModalFooter = $deleteErrorModal.find('.modal-footer'),
```
mbucknell: This could be removed as I don't think it's used.

mhines-usgs:
I like that they can only delete orphaned items, that gives an extra layer of warning. 

I just thought of something looking at your PR, currently you can have the same child item in multiple aggregates, if you orphan one from the tree, does it remove the reference to it if it exists elsewhere or does it automatically delete them from the other aggregates too?

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/tree/UI.js
```diff
@@ -234,17 +234,21 @@ CCH.Objects.Publish.Tree.UI = function (args) {
...
-				if (node.state.displayed) {
-					$('#' + node.li_attr.id + '_anchor').removeClass(invisClass);
-				} else {
-					$('#' + node.li_attr.id + '_anchor').addClass(invisClass);
+		if(allItems !== undefined){
+			for (var cIdx = 0; cIdx < allItems.length; cIdx++) {
+				var node = tree.get_node(allItems[cIdx]);
+				if (node.parent && node.parent !== '#' && node.parent !== 'uber' && node.parent !== 'root') {
+					if (node.state.displayed) {
+						$('#' + node.li_attr.id + '_anchor').removeClass(invisClass);
```
mbucknell: Consider putting the selected element into a variable.

zmoore-usgs:
Ah I didn't think of that test case. I'll try it out and see what happens

mhines-usgs:
me neither! I can make another task if we want to have a separate way to handle those deletes, or we could chat with the cooperators.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
zmoore-usgs: That error is representing that the Item does exist, but it is still nested under some part of the user-visible tree and thus shouldn't be deleted until it has been removed from the visible tree and into the orphan tree.

PR #1616, opened 2017-12-20
CCHS-423: Updated docker files

zmoore-usgs:


mhines-usgs:
Grabbing this to check out!

mhines-usgs:
This is working great!

PR #1617, opened 2018-01-02
CCHS-429 / CCHS-439: Added storm creation from the UI

zmoore-usgs:


coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -2420,6 +2434,193 @@ CCH.Objects.Publish.UI = function () {
...
+		if(copyType == "item") {
+			if(copyText != null && copyText != "") {
+				$.ajax({
+					url: CCH.baseUrl + "/data/item/" + copyText.trim(),
+					type: 'GET',
+					success: function() {
+						me.createNewStorm(alias, copyType, copyText);
+					},
+					error: function() {
+						errorString = "No Item could be retireved using the provided copy from Item ID.";
```
kmschoep-usgs: typo in 'retireved'.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/StormUtil.java
```diff
@@ -0,0 +1,420 @@
...
+        Bbox stormBbox = new Bbox();
+        stormBbox.setBbox(-101.8, 17.6, -62.4, 46.0);
+        return stormBbox;
+    }
+
+    private static List<Service> trackChildServices(String serviceParam) {
+        List<Service> childServices = new ArrayList<>();
+
+        Service sourceWms = new Service();
+        sourceWms.setEndpoint("http://nowcoast.noaa.gov/arcgis/services/nowcoast/wwa_meteocean_tropicalcyclones_trackintensityfcsts_time/MapServer/WmsServer?");
```
mhines-usgs: is it possible at all to not have the storm track as a part of the code? just in case the NHC NOAA makes some kind of change, it might be nice not to have to do a new software release. I'm not sure if that's possible though. I could make a future ticket if that would be good to just separate out from this task.

mhines-usgs: Yeah, I understand why you did it here this way. We had a cheater method of just using the saveTrack() function in hazardItems as needed with the accompanying yaml file.. I don't know what kind of analogous option might exist in java-land, but it's worth exploring. or maybe even managing in the DB instead.. not sure. I'll make a ticket and we can discuss in the future!

mhines-usgs: Created CCHS-444

PR #1619, opened 2018-01-04
CCHS-443: Added the ability to search the item tree by Item ID and Alias

zmoore-usgs:


mhines-usgs:
pulling this locally to check out! :whale:

PR #1623, opened 2018-04-09
Switched example compose.env to use service names instead of IPs.

cschroed-usgs:


compose.env
```diff
@@ -14,9 +14,9 @@ POSTGRES_LOAD_BOOTSTRAP=true
 # Portal
 PORTAL_HTTP_PORT=8080
 PORTAL_HTTPS_PORT=8443
-PORTAL_HOST=192.168.99.100
```
cschroed-usgs: I intentionally changed this to `localhost` instead of `cch_portal`. Something about how we've coded the portal makes the web browser attempt to fetch files from `PORTAL_HOST`. Since developers' web browsers are generally not in a container, it is not using Docker's networking, and it therefore can't resolve the `cch_portal` service name. `localhost` works while running Docker natively. Docker machine users would have to change this to their Docker machine IP

cschroed-usgs:
Don't merge this yet. Running into problems during the RTS import.

cschroed-usgs:
I wasn't able to get this running using solely service names.

PR #1625, opened 2018-04-12
Improved logging and exception message

cschroed-usgs:


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/SLDResource.java
```diff
@@ -110,7 +113,8 @@ public Response getSLDInfo(@PathParam("id") String id, @QueryParam("ribbon") Int
 			}
 		}
 	    } catch(Exception e){
-		    response = Response.status(500).build();
+		LOG.error("Error while getting SLD info", e);
```
cschroed-usgs: Do you know if this should this be at error level or warn level?

zmoore-usgs: I'm not completely sure but error is probably fine. I think we have calls in the UI that hit this when they shouldn't so those may not really be errors but I think it's fine to treat them as such because they should maybe be cleaned up a bit someday.

cschroed-usgs: Sounds good. Thanks!

PR #1626, opened 2018-04-18
CCHS 446 and CCHS 447

cschroed-usgs:
This PR includes new features for:
 * hoisting active storms to uber-level
 * updating an already-published alias to use the new storm and orphaning the previous storm


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/ItemManager.java
```diff
@@ -578,6 +583,80 @@ private boolean isEmpty(List<String> args) {
 		return items;
 	}
 	
+	/**
+	 * Orphans an item.
+	 * 
+	 * Transactionally finds all of this item's ancestors, removes the 
+	 * item from the ancestors' children and merges them back into the DB.
+	 *
```
cschroed-usgs: Do you know if there is a simpler / existing way to accomplish this?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Alias.java
```diff
@@ -51,4 +52,33 @@ public static Alias fromJSON(String json) {
 		
 		return alias;
 	}
+
```
cschroed-usgs: How do you feel about IDE-generated hashCode and equals implementations?
How do you feel about [Project Lombok](https://projectlombok.org/)?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
```diff
@@ -95,7 +96,7 @@
 	/* Whether to show this item at all, used in mediation */
 	private transient boolean enabled;
 	private Summary summary;
-	private List<Service> services;
+	private List<Service> services = new ArrayList<>();
```
cschroed-usgs: Initializing this helps an object 'cloned' via `copyValues(obj, null)` equal its clone according to the `.equals` method.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
```diff
@@ -210,7 +211,7 @@ public void setSummary(Summary summary) {
 	public List<Service> getServices() {
 		return services;
 	}
-
+
```
cschroed-usgs: This pattern of setting the instance variable to an empty list if the user passes in is not something that I usually see on a POJO. Comments on other mutators indicate that we are doing this to support a particular Gson serialization. I'm far from a pro with Gson. Do you know if there is a way we could move these serialization concerns to <ClassName>Adapter classes in order to make the mutators a little more vanilla?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
```diff
@@ -427,4 +439,94 @@ private OGCService fetchOgcService(ServiceType type) {
 		OGCService ogc = Service.ogcHelper(type, services);
 		return ogc;
 	}
+
```
cschroed-usgs: More IDE-generated methods...

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/summary/Summary.java
```diff
@@ -106,21 +106,24 @@ public void setKeywords(String keywords) {
 	}
 
 	public static Summary copyValues(final Summary from, final Summary to) {
```
cschroed-usgs: I found myself using `copyValues` methods in the same way that I would use the more idiomatic `clone()` methods. The `copyValues` method is subtly different from a `clone()` implementation, but I'm wondering if you think it would be worthwhile to build separate `clone` methods?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -142,105 +139,168 @@ public Response instantiateTemplate(@Context HttpServletRequest request, @PathPa
...
+		}
+	}
+	
+	protected Response instantiateStormAndHandleResponse(String templateId, String childJson, String alias, boolean active, AliasManager aliasMan, ItemManager itemMan){
 		Response response;
+		if (StringUtils.isEmpty(templateId)) {
+			//this is the server's fault rather than the client's
+			response = Response.status(500).build();
+		} else {
+			response = instantiateTemplate(templateId, childJson);
```
cschroed-usgs: I broke the original implementation of this method into several smaller methods to permit some testing and to move it towards a more S.L.A.P.-happy approach. (Single Level of Abstraction).

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -142,105 +139,168 @@ public Response instantiateTemplate(@Context HttpServletRequest request, @PathPa
...
 			}
 		}
 		return response;
 	}
 
-	@GET
-	@Path("/storm")
-	@Produces(MediaType.APPLICATION_JSON)
-	@RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
-	public Response instantiateStormTemplate(@Context HttpServletRequest request, @QueryParam("layerId") String layerId, @QueryParam("activeStorm") String active, @QueryParam("alias") String alias, @QueryParam("copyType") String copyType, @QueryParam("copyVal") String copyVal, @QueryParam("trackId") String trackId) {
```
cschroed-usgs: I broke the original implementation of `instantiateStormTemplate` into several smaller methods to permit some testing and to move it towards a more S.L.A.P.-happy approach. (Single Level of Abstraction).

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -142,105 +139,168 @@ public Response instantiateTemplate(@Context HttpServletRequest request, @PathPa
...
+		return response;
+	}
 	
-							if(templateId != null && templateId.length() > 0) {
-								response = instantiateTemplate(request, templateId, childJson);
-								
-								if(response.getStatus() == HttpStatus.SC_OK) {
-									Map<String, Object> ok = new HashMap<String, Object>() {
-										private static final long serialVersionUID = 2398472L;
+	protected void hoistNewTemplateToTopLevel(String newTemplateId, ItemManager itemMan) {
```
cschroed-usgs: This and other semi-awkward error methods that mostly delegate to manager classes are a result of the convention to return booleans rather than throwing Exceptions in the manager classes.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -2725,31 +2724,32 @@ CCH.Objects.Publish.UI = function () {
...
 				$result.append("Failed to post NHC Track Aggregate Item. Storm creation aborted. Error code: " + data.status);
 				$newStormLayerId = null;
 				me.enableNewStormButtons();
 			}
 		});
-	}
-
-	me.postStormTemplate = function($result, layerId, newAlias, copyType, copyText, trackId) {
+	};
+
```
cschroed-usgs: Switched this from callback to a promise-based approach.

cschroed-usgs:
I have hooked up Codacy to CCH to give us better insight on our code quality. Take a peek at the analysis!

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/jpa/ItemManager.java
cschroed-usgs: Ok. Thanks for confirming!

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Alias.java
cschroed-usgs: Cool! I'm not pushing for adopting Project Lombok right now, just hoping to start the conversation. Maybe after we stabilize things for hurricane season we can revisit. More food for thought:
https://stackoverflow.com/questions/4589184/what-are-the-risks-with-project-lombok

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Item.java
cschroed-usgs: For sure -- it would take a while to do the analysis to see if there is a consistent pattern across all models.  I'm weighing some competing concerns on the benefits of a more incremental approach.

Right now both our server-side business logic and our client-side logic are coupled to these quirks. I /think/ (too 'new' to the codebase to say "know" ;) ) that moving the quirks to a layer specifically dedicated to (de)serialization would be better because then only the client side is coupled to the quirks. A downside to an incremental approach is that once we have written some type-specific adapter logic, we might miss patterns that apply to multiple types and wind up with some redundant (de)serialization logic.

Is there anything I'm missing? What do you think about incrementally moving some of the quirks to the Adapters? Not necessarily now, but in general.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/summary/Summary.java
cschroed-usgs: Cool! Sounds like we're on the same page.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
cschroed-usgs: Yeah. I don't understand hibernate or CCH-specific stuff enough yet to criticize the Manager convention of returning booleans, just noting that it resulted in some weirdness in this case.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -264,18 +324,18 @@ private Summary copyExistingSummary(String copyType, String copyVal, ItemManager
 		Summary newSummary = null;
 		Item summaryItem = null;
 
-		if(copyType.equalsIgnoreCase("item")) {
+		if ("item".equalsIgnoreCase(copyType)) {
 			summaryItem = itemMan.load(copyVal);
-		} else if(copyType.equalsIgnoreCase("alias")) {
+		} else if ("alias".contains(copyType)) {
```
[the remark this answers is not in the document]
cschroed-usgs: Good catch! Will fix.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -142,105 +139,168 @@ public Response instantiateTemplate(@Context HttpServletRequest request, @PathPa
...
-							
-							Item baseTemplate = baseTemplateItem(Boolean.parseBoolean(active), layer.getBbox(), serviceCopies, summary);
-							String templateId = itemMan.persist(baseTemplate);
+			if (response.getStatus() == HttpStatus.SC_OK) {
+				if (active) {
+					hoistNewTemplateToTopLevel(templateId, itemMan);
+				}
+				if (!StringUtils.isEmpty(alias)) {
+					String originalTemplateId = updateStormAlias(alias, aliasMan, templateId);
+					orphanOriginalTemplate(originalTemplateId, itemMan);
```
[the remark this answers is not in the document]
mhines-usgs: Not sure I follow what use case you mean?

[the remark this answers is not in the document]
cschroed-usgs: 3 and 4 are implemented now. I was able to add some tests for 3. I can't test 4 without some nontrivial refactoring.

PR #1628, opened 2018-04-19
Updated Hurricane legend icons

mwernimont:
Updated the colors of our current hurricane legend icons to match NOAA's updated icons.


mhines-usgs:
pulling it locally to check out! thanks!

PR #1629, opened 2018-04-19
Caching maven dependencies to speed up build.

cschroed-usgs:


cschroed-usgs:
https://docs.travis-ci.com/user/caching/#Arbitrary-directories

PR #1632, opened 2018-04-20
Automatically Report Test Coverage

cschroed-usgs:
Due to pom changes in #1631 , there are now HTML, CSV, and XML coverage report files in the `target` directory of each maven submodule. If IDEs support viewing jacoco coverage, they should pick up on the report files after a build and use them to paint our lines of code red, yellow, or green according to their coverage status.

This pull request adds a script that finds all XML coverage reports and uploads them to Codacy. The script only performs the reporting if the appropriate security tokens are present in the environment variables. We associate the security tokens with secret environment variables in the [USGS-CIDA repo's Travis setting page](https://travis-ci.org/USGS-CIDA/coastal-hazards/settings). Secret env vars are not available to pull requests issued from forks because someone could just PR some code that echoes the secrets and then go have a lot of fun with them :grinning:  Secret env vars _are_ available to post-merge builds and to pull requests issued between branches within the canonical (USGS-CIDA) repository.

Since we currently use a fork-and-feature-branch git workflow, we can't take advantage of coverage analysis in pull requests. Would it be cool to try out a feature branch workflow that doesn't use forks to see if the coverage info is helpful at pull request time?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ jdk:
 cache:
   directories:
     - $HOME/.m2
+    - $HOME/codacy
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: java
 jdk:
   - oraclejdk8
+
+after_success:
+  - ./reportCoverageToCodacy.sh
+
 cache:
   directories:
     - $HOME/.m2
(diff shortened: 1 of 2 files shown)


cschroed-usgs:
I'm glad you're excited about this too! Let's try it out. If it ends up being helpful we can document and share.

PR #1634, opened 2018-04-24
Troubleshooting code coverage reporting

cschroed-usgs:


.travis.yml
```diff
@@ -1,7 +1,10 @@
 language: java
+
 jdk:
   - oraclejdk8
 
+sudo: false
```
cschroed-usgs: opt in to fast container-based builds

reportCoverageToCodacy.sh
```diff
@@ -11,10 +11,13 @@ then
 		#if a copy of the code reporter has not already been cached
 		if [ ! -f ~/codacy/coverage-reporter-assembly-latest.jar ]
 		then
-			#install a tool for parsing a REST API response from GitHub
-			sudo apt-get install jq
-			#install the latest release from GitHub
-			wget -O ~/codacy/coverage-reporter-assembly-latest.jar $(curl https://api.github.com/repos/codacy/codacy-coverage-reporter/releases/latest | jq -r .assets[0].browser_download_url)
```
cschroed-usgs: This dynamic detection of the most recent release required installing a new OS-level package. This produced significantly slower build times (~4-5x slower). I switched to retrieving reporting jar from a static url so that we could improve build times.

PR #1638, opened 2018-05-02
CCHS 432

cschroed-usgs:


util/layer-sweep/find_orphaned_layers.py
```diff
@@ -0,0 +1,169 @@
...
+		}
+		return args
+
+def main(argv):
+	parsed_args = parse_cmd_args(argv)
+	orphaned_layers = find_orphaned_layers(**parsed_args)
+	orphaned_layer_names = '\n'.join([ l.name for l in orphaned_layers ])
+	return orphaned_layer_names
+
+if '__main__' == __name__ :
```
cschroed-usgs: @zmoore-usgs this is the entry point, a good spot to start review.

util/layer-sweep/find_orphaned_layers.py
```diff
@@ -0,0 +1,169 @@
...
+	command	are sent to stdout
+	"""
+	print(*args, file=sys.stderr, **kwargs)
+
+def get_cch_items(cch_url):
+	"""
+	returns an iterable of dicts representing items that CCH is aware of
+	"""
+	#'subtree=false' produces a flat list of all items
+	all_items_url = cch_url + 'data/item?subtree=false&showDisabled=true'
```
cschroed-usgs: Should `showDisabled` be set to true?

util/layer-sweep/find_orphaned_layers.py
```diff
@@ -0,0 +1,169 @@
...
+	has_cch_geoserver_services = 0 != len(cch_geoserver_services)
+	is_data = 'data' == item['itemType']
+	return is_data and has_cch_geoserver_services;
+
+def get_only_cch_geoserver_services(services):
+	"""
+	Filters the parameterized list of services (dicts). The returned list 
+	should only contain services (dicts) that have `endpoint`s that look
+	like they are from a cch geoserver.
+	"""
```
cschroed-usgs: The filter criteria in this function is a likely source of bugs. I arrived at these criteria by analyzing the unique 'endpoint's of all of services of all of the items on all three tiers. Please let me know if anything looks amiss.

util/layer-sweep/find_orphaned_layers.py
zmoore-usgs: I believe so. From my recollection disabled items are only "hidden" from the UI but could be re-enabled at any time so the layers for them in GeoServer still need to exist.

util/layer-sweep/find_orphaned_layers.py
```diff
@@ -0,0 +1,169 @@
...
+	all_gs_layers = get_geoserver_layers(geoserver_url, geoserver_username, geoserver_password)
+	
+	cch_items = get_cch_items(cch_url)
+	registered_geoserver_layer_names = convert_cch_items_to_cch_geoserver_layer_names(cch_items)
+	orphaned_gs_layers = filter_out_geoserver_layers_that_are_registered(all_gs_layers, registered_geoserver_layer_names)
+	eprint(
+		("\nTotal Layers Retrieved From GeoServer: {}\n" +
+		"Total Items Retrieved From CCH Portal: {}\n" +
+		"Total Orphaned Items: {}\n").format(
```
zmoore-usgs: Just a bit of a nitpick, but perhaps this should be "Total Orphaned Layers" instead just to prevent confusion that it might be referring to Orphaned Items from CCH?

util/layer-sweep/find_orphaned_layers.py
zmoore-usgs: The filter criteria seem reasonable to me. I can't think of anything that would be an exception to any of these.

util/layer-sweep/find_orphaned_layers.py
cschroed-usgs: Sweet. Thanks for confirming!

util/layer-sweep/find_orphaned_layers.py
cschroed-usgs: That is a great clarification. Will do. Thank you!

PR #1640, opened 2018-05-08
Increased max memory allocation pool for maven-surfire which is what…

zmoore-usgs:
…includes jacoco which is where the memory issue seems to be occurring on Jenkins.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -486,7 +486,7 @@
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.19.1</version>
 				<configuration>
-					<argLine>-Xmx1024m ${surefireArgLine}</argLine>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -486,7 +486,7 @@
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.19.1</version>
 				<configuration>
-					<argLine>-Xmx1024m ${surefireArgLine}</argLine>
+					<argLine>-Xmx2048m ${surefireArgLine}</argLine>
```
cschroed-usgs: Nice find! It's kinda funny that allocating more RAM to this step helps the Jenkins memory issue. Stranger things have happened in the wild world of software development though :smile:

PR #1643, opened 2018-06-28
Sugar Free CCH

zmoore-usgs:
- Removed SugarJS: It was pretty heavily used however many of the features we were using have since been added (or had equivalent functionality added) to base JavaScript and after going through:  https://github.com/andrewplummer/sugar-object/blob/master/CAUTION.md it looked like upgrading would be just about as much work as just ripping it out.

Due to the way things in Sugar extended base objects and in some cases overloaded the names of base JS functions it was very difficult to find everything so I'm not absolute sure at this point that I've found all of the pieces that need to get ripped out, but at this point I'm at least not seeing any errors and I'm getting the behavior I expect (at least in everything I've tested so far).

Some things are made even more difficult because the behavior can vary based on the underlying object being operated upon. For example, both jQuery and SugarJS add a `.last()` method for objects. jQuery's works only on jQuery objects, and Sugar's works on Arrays. In some cases we didn't follow consistent naming conventions and where most jQuery variables are named with a `$` in front, some weren't so I replaced their `.last()` call with `.slice(-1)` when I shouldn't have. I think I've corrected those, but it's possible there are still others I've missed.

I'm going to continue mucking around in this a bit more between Sprint 0 meetings but I wanted to get it up for review early because it's starting to sprawl a bit.

### DIFF
--- a/coastal-hazards-portal/pom.xml
+++ b/coastal-hazards-portal/pom.xml
@@ -218,10 +218,6 @@
 		</dependency>
 		<dependency>
 			<groupId>org.webjars</groupId>
-			<artifactId>sugar</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.webjars</groupId>
 			<artifactId>openlayers</artifactId>
 		</dependency>
 		<dependency>
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
@@ -87,11 +87,17 @@ CCH.Objects.Item = function (args) {
 			// If I have children, load those as well
 			var setLoaded = function (evt, args) {
 				if (!me.loaded) {
-					var loadedItemIsChild = me.children.findIndex(function (childId) {
-						return childId === args.id;
-					}) !== -1,
-							childItems = [],
-							allLoaded;
+					var loadedItemChildIndex = -1;
+					me.children.some(function (childId, index) {
+						if(childId === args.id) {
+							loadedItemChildIndex = index;
+							return true;
+						}
+					});
+
+					var loadedItemIsChild = loadedItemChildIndex !== -1, 
+						childItems = [],
+						allLoaded;
 					if (loadedItemIsChild) {
 						me.children.forEach(function (childId) {
 							var childItem = CCH.items.getById({id: childId});
@@ -101,9 +107,14 @@ CCH.Objects.Item = function (args) {
 						});
 
 						if (childItems.length === me.children.length) {
-							allLoaded = childItems.findIndex(function (childItem) {
-								return !childItem.loaded;
-							}) === -1;
+							var unloadedIndex = -1;
+							childItems.some(function (childItem, index) {
+								if(!childItem.loaded) {
+									unloadedIndex = index;
+									return true;
+								}
+							});
+							allLoaded = unloadedIndex === -1;
 							if (allLoaded) {
 								me.loaded = true;
 								CCH.LOG.debug('Item.js::init():Item ' + me.id + ' finished initializing.');
@@ -251,8 +262,12 @@ CCH.Objects.Item = function (args) {
 			}
 		} else {
 			if (me.ribboned && me.parent && me.parent.ribboned) {
-				index = me.parent.children.findIndex(function (childId) {
-					return me.id === childId;
+				index = -1;
+				me.parent.children.some(function (childId, indx) {
+					if(me.id === childId) {
+						index = indx;
+						return true;
+					}
 				});
 				index += 1;
 				if (index > layers.length + 1) {
@@ -339,8 +354,12 @@ CCH.Objects.Item = function (args) {
 			layerName = aggregationName + this.id;
 
 			if (me.ribboned && me.parent && me.parent.ribboned) {
-				index = me.parent.children.findIndex(function (childId) {
-					return me.id === childId;
+				index = -1;
+				me.parent.children.some(function (childId, indx) {
+					if(me.id === childId) {
+						index = indx;
+						return true;
+					}
 				});
 				index += 1;
 
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
@@ -93,7 +93,7 @@ CCH.Objects.Item = function (args) {
 							childItems = [],
 							allLoaded;
 					if (loadedItemIsChild) {
-						me.children.each(function (childId) {
+						me.children.forEach(function (childId) {
 							var childItem = CCH.items.getById({id: childId});
 							if (childItem) {
 								childItems.push(childItem);
@@ -154,7 +154,7 @@ CCH.Objects.Item = function (args) {
 				loadDataToItem(me, null);
 			} else {
 				// The child still needs to be loaded
-				me.children.each(function (child) {
+				me.children.forEach(function (child) {
 					new CCH.Objects.Item({
 						id: child,
 						parent: me
@@ -379,7 +379,7 @@ CCH.Objects.Item = function (args) {
 	};
 
 	me.hideLayer = function () {
-		me.getLayerList().layers.each(function (layerName) {
+		me.getLayerList().layers.forEach(function (layerName) {
 			CCH.map.hideLayersByName(layerName);
 		});
 
@@ -471,14 +471,16 @@ CCH.Objects.Item = function (args) {
 		var children = args.children || [],
 				item = args.item || me,
 				itemChildren = item.children;
-				children.add(item);
+				children.push(item);
 				
 		if (itemChildren.length) {
 			for (var cIdx = 0; cIdx < itemChildren.length; cIdx++) {
 				var child = CCH.items.getById({ id : itemChildren[cIdx] });
-				children.union(me.getChildren({
+				children = children.concat(me.getChildren({
 					children : children,
 					item : child
+				}).filter(function(childItem) {
+					return children.indexOf(childItem) < 0;
 				}));
 			}
 		}
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
@@ -87,15 +87,9 @@ CCH.Objects.Item = function (args) {
 			// If I have children, load those as well
 			var setLoaded = function (evt, args) {
 				if (!me.loaded) {
-					var loadedItemChildIndex = -1;
-					me.children.some(function (childId, index) {
-						if(childId === args.id) {
-							loadedItemChildIndex = index;
-							return true;
-						}
-					});
-
-					var loadedItemIsChild = loadedItemChildIndex !== -1, 
+					var loadedItemIsChild = me.children.some(function (childId) {
+							return childId === args.id
+						}),
 						childItems = [],
 						allLoaded;
 					if (loadedItemIsChild) {
@@ -107,14 +101,9 @@ CCH.Objects.Item = function (args) {
 						});
 
 						if (childItems.length === me.children.length) {
-							var unloadedIndex = -1;
-							childItems.some(function (childItem, index) {
-								if(!childItem.loaded) {
-									unloadedIndex = index;
-									return true;
-								}
+							var allLoaded =	childItems.every(function (childItem) {
+								return childItem.loaded;
 							});
-							allLoaded = unloadedIndex === -1;
 							if (allLoaded) {
 								me.loaded = true;
 								CCH.LOG.debug('Item.js::init():Item ' + me.id + ' finished initializing.');
@@ -262,18 +251,6 @@ CCH.Objects.Item = function (args) {
 			}
 		} else {
 			if (me.ribboned && me.parent && me.parent.ribboned) {
-				index = -1;
-				me.parent.children.some(function (childId, indx) {
-					if(me.id === childId) {
-						index = indx;
-						return true;
-					}
-				});
-				index += 1;
-				if (index > layers.length + 1) {
-					index = layers.length + 1;
-				}
-
 				if (bboxObject[stringifiedBbox]) {
 					index = bboxObject[stringifiedBbox].length + 1;
 					bboxObject[stringifiedBbox].push(me.id);
@@ -354,19 +331,6 @@ CCH.Objects.Item = function (args) {
 			layerName = aggregationName + this.id;
 
 			if (me.ribboned && me.parent && me.parent.ribboned) {
-				index = -1;
-				me.parent.children.some(function (childId, indx) {
-					if(me.id === childId) {
-						index = indx;
-						return true;
-					}
-				});
-				index += 1;
-
-				if (index > layers.length + 1) {
-					index = layers.length + 1;
-				}
-
 				if (bboxObject[stringifiedBbox]) {
 					index = bboxObject[stringifiedBbox].length + 1;
 					bboxObject[stringifiedBbox].push(me.id);
@@ -498,9 +462,9 @@ CCH.Objects.Item = function (args) {
 				children = children.concat(me.getChildren({
 					children : children,
 					item : child
-				}).filter(function(childItem) {
-					return children.indexOf(childItem) < 0;
-				}));
+				})).filter(function(childItem, index, self) {
+					return index == self.indexOf(childItem);
+				});
 			}
 		}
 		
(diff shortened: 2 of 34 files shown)


mhines-usgs:
I pulled this to try and be helpful/identify any issues, but it fails right away for me on loading the public facing portal with an error on the loading page ("There was a problem loading the page. Please try to reload or contact us"), in the console I see 
```
Uncaught TypeError: $parentContainerVisibleItems.toArray(...).sum is not a function
    at me.resizeContainer (CombinedSearchbar.js:70)
    at dispatch (jquery.js:4409)
    at elemData.handle (jquery.js:4095)
    at Object.trigger (jquery.js:4324)
    at jquery.js:4875
    at Function.each (jquery.js:375)
    at jQuery.fn.init.each (jquery.js:139)
    at jQuery.fn.init.trigger (jquery.js:4874)
    at Object.me.windowResizeHandler (UI.js:142)
    at UI.js:565
```

zmoore-usgs:
Huh.. Interesting mine starts up fine. I'll take a look

zmoore-usgs:
Hmm perhaps that line is being cached for you or the rebuild you're running doesn't include these updates? That function it's referencing should be removed..

https://github.com/USGS-CIDA/coastal-hazards/pull/1643/files#diff-a9ddbde4f8c701248f3e2ecb8fba1cfaL70

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -267,8 +267,8 @@ return {
 					height;
 				
 				if (layerName.indexOf('_r_') !== -1) {
-					layerParts = layerName.split('_');
-					layerPart = layerParts[layerParts.length-1];
+					var layerParts = layerName.split('_'),
+						layerPart = layerParts[layerParts.length-1];
```
mbucknell: I prefer declaring each variable using a separate var. It's more maintainable and you won't accidently create global variables when deleting lines.

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
```diff
@@ -87,23 +87,34 @@ CCH.Objects.Item = function (args) {
 			// If I have children, load those as well
 			var setLoaded = function (evt, args) {
 				if (!me.loaded) {
-					var loadedItemIsChild = me.children.findIndex(function (childId) {
-						return childId === args.id;
-					}) !== -1,
-							childItems = [],
-							allLoaded;
+					var loadedItemChildIndex = -1;
+					me.children.some(function (childId, index) {
```
mbucknell: I think you could simplify this by the following:
```
var loadedItemIsChild = me.children.some(function(childId, Index)
```
Then you don't need to declare loadedItemChildIndex

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
```diff
@@ -87,23 +87,34 @@ CCH.Objects.Item = function (args) {
...
 								childItems.push(childItem);
 							}
 						});
 
 						if (childItems.length === me.children.length) {
-							allLoaded = childItems.findIndex(function (childItem) {
-								return !childItem.loaded;
-							}) === -1;
+							var unloadedIndex = -1;
+							childItems.some(function (childItem, index) {
```
mbucknell: Same simplification as above should work here, I think.

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
```diff
@@ -251,8 +262,12 @@ CCH.Objects.Item = function (args) {
...
-					return me.id === childId;
+				index = -1;
+				me.parent.children.some(function (childId, indx) {
+					if(me.id === childId) {
+						index = indx;
+						return true;
+					}
 				});
 				index += 1;
 				if (index > layers.length + 1) {
```
mbucknell: I don't understand what is going on in lines 273 and below. index is reassigned in the if clause starting at 277 so I don't know what the code from lines 265 through 277 accomplish

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
```diff
@@ -339,8 +354,12 @@ CCH.Objects.Item = function (args) {
 			layerName = aggregationName + this.id;
 
 			if (me.ribboned && me.parent && me.parent.ribboned) {
-				index = me.parent.children.findIndex(function (childId) {
-					return me.id === childId;
+				index = -1;
```
mbucknell: Same thing is going on here as in above comment.

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -579,15 +579,18 @@ CCH.Objects.Publish.UI = function () {
...
-				.not(':first')
-				.toArray()
-				.count(function (input) {
-					return $(input).val().trim() === keyword.trim();
-				}) > 0;
+		$form.find('.form-group-keyword input')
+			.not(':first')
+			.toArray()
+			.forEach(function (input) {
+				if($(input).val().trim() === keyword.trim()) {
```
mbucknell: nit. Could assign the result of keyword.trim() to variable to avoid repeated calls.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/BucketSlide.js
```diff
@@ -352,9 +352,14 @@ CCH.Objects.Widget.BucketSlide = function (args) {
 	};
 
 	me.getCardIndex = function (id) {
-		return me.cards.findIndex(function (i) {
-			return i.data('id') === id;
+		var cIndex = -1;
+		me.cards.some(function (i, index) {
```
mbucknell: Have you tried this on IE11? I have a link to a page that says this won't work on IE11. http://www.angulartutorial.net/2017/03/findIndex-issue-in-IE.html

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/BucketSlide.js
```diff
@@ -437,9 +446,13 @@ CCH.Objects.Widget.BucketSlide = function (args) {
 			// the bucket class will actually call this function with a proper
 			// id. It's a long way around removing the item but it does hit 
 			// multiple components
-			me.cards.reverse().each(function ($card) {
+			var deleteList = [];
+			me.cards.reverse().forEach(function ($card) {
```
mbucknell: You might consider using .map rather than .forEach

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/BucketSlide.js
```diff
@@ -584,10 +598,15 @@ CCH.Objects.Widget.BucketSlide = function (args) {
 			$imageContainer = $cardHtml.find('.application-slide-bucket-container-card-image');
 
 		// Test if the layer is currently visible. If not, set view button to off 
-		layerCurrentlyInMap = item.getLayerList().layers.every(function (id) {
+		layerCurrentlyInMap = true;
+		item.getLayerList().layers.some(function (id) {
```
mbucknell: Consider simplifying by assigning layerCurrentlyInMap to the result of this line (with a ! to negate)

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
zmoore-usgs: I'll take a look into these and see if I can figure out what's going on. Didn't really pay much attention to what the code was actually doing in this PR outside just mimicking the functionality that SugarJS was providing.

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
zmoore-usgs: Ah okay, sounds good; I prefer it that way as well I just did it this way because I was getting warnings during compile about having more than one `var` per scope.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/BucketSlide.js
zmoore-usgs: I haven't tried it, I just went by this which lists support as of IE 9:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

mbucknell: The .some function is supported, I think it's the part where you are assigning the index to a variable outside the scope of the .some function.

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
mbucknell: How are you compiling the code?Are you compiling the javascript? If not, it's probably the linter that's been configured.

coastal-hazards-portal/src/main/webapp/js/cch/objects/widget/BucketSlide.js
zmoore-usgs: Ah, I see okay. I'll look into converting to just a basic for loop then.

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
zmoore-usgs: Yeah, it's a linter that runs during the maven build

coastal-hazards-portal/pom.xml
```diff
@@ -216,10 +216,6 @@
 			<groupId>org.webjars</groupId>
 			<artifactId>jquery</artifactId>
 		</dependency>
-		<dependency>
```
cschroed-usgs: Could you please remove the corresponding dependency entry from the parent pom as well?

cschroed-usgs: Could you also please remove the sugar js version property from the parent pom?

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
```diff
@@ -471,14 +490,16 @@ CCH.Objects.Item = function (args) {
...
 				
 		if (itemChildren.length) {
 			for (var cIdx = 0; cIdx < itemChildren.length; cIdx++) {
 				var child = CCH.items.getById({ id : itemChildren[cIdx] });
-				children.union(me.getChildren({
+				children = children.concat(me.getChildren({
 					children : children,
 					item : child
+				}).filter(function(childItem) {
+					return children.indexOf(childItem) < 0;
```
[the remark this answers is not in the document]
cschroed-usgs: Is this a correct replacement for sugar's .union method?

> Returns a new array containing elements in both arrays with duplicates removed. In addition to primitives, this method will also deep-check objects for equality.

-- https://sugarjs.com/docs/#/Array/union

I tried to create a simple example of this technique.
http://jsbin.com/setotisefi/2/edit?html,js,console,output
It shows they produce different results. Is the example wrong, or is the workaround?

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
```diff
@@ -350,16 +350,20 @@ CCH.Objects.Back.UI = function (args) {
 				if (item.children.length !== 0) {
 					for (var cIdx = 0; cIdx < item.children.length; cIdx++) {
 						var childId = item.children[cIdx];
-						services = services.union(deepDiveToFindAnyServices(CCH.items.getById({id: childId}), services));
```
cschroed-usgs: Same question: Is this a correct replacement of sugar's union method?

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
```diff
@@ -350,16 +350,20 @@ CCH.Objects.Back.UI = function (args) {
...
+						services = services.concat(deepDiveToFindAnyServices(CCH.items.getById({id: childId}), services).filter(function(service) {
+							return services.indexOf(service) < 0;
+						}));
 					}
 				}
-				services = services.union(item.services);
```
cschroed-usgs: And here

coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/tree/UI.js
```diff
@@ -491,10 +491,14 @@ CCH.Objects.Publish.Tree.UI = function (args) {
 			if (displayed) {
 				node.state.displayed = false;
 				$('#' + selectedId + '_anchor');
-				parent.state.displayedChildren.remove(originalId);
+				parent.state.displayedChildren = parent.state.displayedChildren.filter(function(el) {
+					el !== originalId
+				});
 			} else {
 				node.state.displayed = true;
-				parent.state.displayedChildren = parent.state.displayedChildren.union(originalId);
```
cschroed-usgs: This workaround looks reasonable because originalId is just one object, not many.

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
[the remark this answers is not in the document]
zmoore-usgs: Yeah, there is a bug here. Rather than checking against `children` for the index I need to check against a new array and add the de-duped items into that. See: http://jsbin.com/leqokidani/edit?html,js,console,output

I'm also following Mary's advice and changing the implementation so this will get fixed along with that.

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
zmoore-usgs: Yeah I really don't know what the point of the code from 265-277 is.. The value set there is not used before it is set again.. I'm going to axe it.

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
zmoore-usgs: Axed this as well

coastal-hazards-portal/src/main/webapp/js/cch/objects/Item.js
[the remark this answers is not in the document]
zmoore-usgs: Actually never mind, I confused myself and didn't look closely. What's written here should theoretically work but referring to the children in the filter sounds like it doesn't always work across browsers as @mbucknell pointed out below.

The code written here works because the filter is happening within the concat, not after as in your jsbin. See:

http://jsbin.com/zulopanura/edit?html,js,console,output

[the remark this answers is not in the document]
cschroed-usgs: Aha! one paren makes all the difference. Thanks!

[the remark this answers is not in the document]
zmoore-usgs: Yeah it fooled me too until I was looking at it second time while running through this feedback list again haha. With so many nested parens and brackets it certainly is difficult to read.

coastal-hazards-portal/src/main/webapp/js/cch/objects/front/UI.js
```diff
@@ -336,8 +342,14 @@ CCH.Objects.Front.UI = function (args) {
 					item: itemById,
 					visibility: sessionItem.visible
 				});
-				me.bucket.bucket = me.bucket.getItems().sortBy(function (i) {
-					return i.addAtIndex;
+				me.bucket.bucket = me.bucket.getItems().sort(function (a, b) {
+					if(a.addAtIndex < b.addAtIndex) {
```
cschroed-usgs: Could these conditions be replaced with something more concise like `a.addAtIndex - b.addAtIndex`?

coastal-hazards-portal/src/main/webapp/js/cch/objects/front/UI.js
```diff
@@ -454,8 +466,14 @@ CCH.Objects.Front.UI = function (args) {
 							if (Object.keys(bucketLegends).length === this.$container.children().length) {
 								// I am the final card that will be loaded. I need to organize my container to 
 								// be indexed in the same way that the bucket is
-								var sortedLegends = this.$container.find('>div').toArray().sortBy(function ($div) {
-									return parseInt($($div).attr('card-index'));
+								var sortedLegends = this.$container.find('>div').toArray().sort(function (a,b) {
```
cschroed-usgs: Would subtraction work instead of conditions?

PR #1653, opened 2018-08-07
Docker cleanup portal

isuftin:


isuftin:
I need to update the rserve section before merging this. I will merge the rserve section to here before merging this PR. This still builds rserve 3.4.3 but i need to get it to 3.3.3

isuftin:
Should be good once https://github.com/USGS-CIDA/coastal-hazards/pull/1654 is merged

PR #1663, opened 2018-09-25
CCH 492: Remove need for PyCsw when publishing/reading metadata.

dpattermann-usgs:
Removed need for PyCsw when reading metadata.
Metadata should be returned as GSON in a map with keys: Keywords, Box, Resources.


zmoore-usgs:
I think those docker keystores are supposed to still be included? I can't do a docker-compose up with the local profile because those are gone, unless we're supposed to generate our own keystores - which I could do.

AQCU and MLR come bundled with keystores for their local development stuff so I think it would be okay for CCH to as well?

dpattermann-usgs:
@zmoore-usgs Does it work for you now?

zmoore-usgs:
Re-pulling and testing now

zmoore-usgs:
All good now, thanks!

dpattermann-usgs:
Wrote a test for extractStringsFromCswDoc to alert us if it breaks. Will merge after checks are made.

dpattermann-usgs:
I cleaned up some of the errors coming from Codacy, going to merge now.

PR #1667, opened 2018-10-12
CCH-494: Remove pycsw

dpattermann-usgs:
After merging CCH 493, this should be easier compare. Stops the startup of pycsw, I believe. Please correct me if this is at all wrong.


compose.env
```diff
@@ -1,4 +1,4 @@
-EXTERNAL_HOST=10.0.2.15
+EXTERNAL_HOST=130.11.170.91
```
zmoore-usgs: May want to change this to a placeholder since whomever runs this will likely need to change it

PR #1670, opened 2019-02-20
Remove csw records via liquibase

mhines-usgs:


mhines-usgs:
Ran this via the docker setup locally and can confirm it ran evidenced by the record in the changelog and the lack of csw records in the service table.

PR #1672, opened 2019-04-23
Update version because of security issue

mhines-usgs:


mhines-usgs:
closing, addressed in https://github.com/USGS-CIDA/coastal-hazards/pull/1673

PR #1678, opened 2019-08-23
Conversion from CIDA Auth to Keycloak and some cleanup

zmoore-usgs:
I included a bunch of cleanup in this PR as well, so the important files to look at for the CIDA Auth to Keycloak conversion are:

1. README.MD & Docker/README.MD
2. Docker/keycloak/*
3. coastal-hazards-portal/Dockerfile & coastal-hazards-portal/Dockerfile.local
4. coastal-hazards-portal/docker/*
5. coastal-hazards-portal/pom.xml
6. coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/*
7. coastal-hazards-portal/src/main/webapp/*
8. compose.env & docker-compose.yml

I also made a fair number of changes to all of the Dockerfiles because I couldn't get them to build correctly within the DOI Network (even with doi_network=true defined), so I fixed the issues I ran into.

The _vast_ majority of the added lines of code are in the `cch_local_development_realm.json` file added for auto configuration of the local development Keycloak docker container. Without the 1800 lines of that file this PR would be a reduction versus an increase.


coastal-hazards-portal/src/main/webapp/js/cch/objects/publish/UI.js
```diff
@@ -1588,7 +1584,7 @@ CCH.Objects.Publish.UI = function () {
 	});
 
 	$buttonLogout.on(CCH.CONFIG.strings.click, function () {
-		CCH.Auth.logout();
+		alert("logout");
```
zmoore-usgs: Still working on this, but wanted to get the PR up to get review started.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/publish/PublishResource.java
```diff
@@ -1,87 +1,74 @@
...
+
+import gov.usgs.cida.coastalhazards.rest.security.ConfiguredRolesAllowed;
 
 /**
  *
- * @author isuftin
+ * @author isuftin, zmoore
  */
 @Path("/")
-@PermitAll //says that all methods, unless otherwise secured, will be allowed by default
```
zmoore-usgs: The reason this looks almost completely rewritten is because I cleaned up some whitespace issues that were present initially. The only real change here is replacing `@PermitAll` at the class-level with my new `@ConfiguredRolesAllowed` annotation because previously every method defined in this class was secured with `@RolesAllowed(<admin role>)` anyways.

Docker/rserve/Dockerfile
```diff
@@ -25,6 +19,13 @@ RUN apt-get update && \
 
 RUN mkdir work
 
+ARG doi_network=false
+
```
isuftin: My suggestion:

Remove the doi_network ARG completely and below, try:

```docker
RUN if getent ahosts sslhelp.doi.net > /dev/null 2>&1; then \
             do whatever
         fi
```

Here's why: sslhelp.doi.net is not resolvable outside of the DOI network. Therefore if the host building the image is able to resolve this host, they are on the doi network. No need to have the argument passed in. `getent` is a utility that is already installed on `tomcat:8.0-jre8-alpine` which is what is being used here. The test above checks whether the host is resolvable, and if so, installs the SSL certs from sslhelp.doi.net

zmoore-usgs: As long as I can get an equivalent thing to work in the other docker images that currently use `doi_network` that seems reasonable to be :+1:

isuftin: I've tested this as also working on:

- mdillon/postgis:9.6-alpine (the liquibase docker image we build) 
- rocker/r-ver:3.3.3 (the rserve image we build) 

Did I miss any others?

zmoore-usgs: I don't think so :+1:

isuftin:
At the top of `Docker/README.MD` there's a line that states:

`# Note this readme is outdated but kept for reference purposes. See the project readme for updated instructions for using docker. If you'd like more information about how the docker files work then read this readme.' (you added this on 12.20.17) -- So should this file continue to be added to?

zmoore-usgs:
I think actually I should remove that notice because it's not really out-of-date at this point. I think we were planning to significantly change the Dockerization again but didn't end up actually doing it.

Docker/README.md
```diff
@@ -69,14 +70,35 @@ If you are not on the doi network, do not specify that variable. If you need par
...
 
 
+#### Keycloak
+=========
+
+CCH Uses Keycloak for authentication into the mediation page, and as such when running in Docker we must have a Keycloak instance available for the portal to authenticate with. The cch_keycloak container is started and automatically configured using the `cch_local_development_realm.json` file in `./Docker/keycloak/`. In order for Keycloak to work properly it must be available on **the same URL** both _within_ and _outside_ of the Docker network. Within the network created by the Docker Compose file the Keycloak container runs on the `keycloak` hostname, however this hostname is not directly available to us outside of the Docker network. To get around this, you should add an entry to your host OS `hosts` file to map the hostname `keycloak` to `127.0.0.1` so that when you open a browser to `keycloak:8083/auth/` you end up at the keycloak site running on `127.0.0.1:8083`. This will allow the OAuth2 login flow to work properly between the Keycloak server and Portal server. This step **must** be done regardless of whether the portal is running in Docker or on a Tomcat instance outside of Docker.
```
isuftin: Reading this in an IDE that does not create virtual linebreaks is a test in patience for horizontal scrolling.

image

I wonder if we can break markdown lines at 80 columns with an actual carriage return?

isuftin: One caveat about the above instructions is if you have a remote Docker engine running (e.g. Docker Machine), it won't be 127.0.0.1 but the ip of docker machine vm.  So perhaps add an extra instruction saying something like:

```
If using a remote Docker engine, substitute 127.0.0.1 with the IP of the host running your Docker engine.

If using Docker Machine, use the name of the VM to find out the ip. For example, if the VM's name is "workbench", issue the following command:

`docker-machine ip workbench`
```

Docker/README.md
```diff
@@ -121,27 +143,14 @@ Once the server is up and running, you can access the server by pointing your we
 
 This container is standalone and does not depend on any other containers to run. You can run the container by issuing:
 
-`docker-compose up cch_db`
+`docker-compose up cch_postgres`
 
 The DOI SSL issue also applies to this container, so if you are on the DOI network, issue:
```
isuftin: These instructions can be removed once the images are doing self-tests for being on the DOI network or not

Docker/README.md
zmoore-usgs: I used a VSCode extension to re-wrap the READMEs to 81 columns (that's what the extension wraps to). It seems to have preserved URL lines to over 81 columns since that's needed for them to work, but otherwise everything else seems to be changed and it still looks fine once rendered.

Docker/README.md
zmoore-usgs: Yep, removed them

Docker/README.md
```diff
@@ -154,22 +163,21 @@ Not seeing the changes you expect? Try building without a cache by using `--buil
 Example:
 
 ```
-CCH_ENV_LOCAL="_johns_laptop" doi_network="true" docker-compose up --build --force-recreate cch_db cch_rserve cch_n52_wps cch_pycsw
+CCH_ENV_LOCAL="_johns_laptop" doi_network="true" docker-compose up --build --force-recreate cch_postgres cch_rserve cch_n52_wps
 ```
 
 ##### DB-related Startup Errors
```
isuftin: Is this still a thing? Do you know why?

I think this may have been related:
https://github.com/docker/compose/pull/3433

This was an issue in Docker Compose 1.7.1 and they're on 1.24.1 now - I think it's safe to remove this at this point.

Docker/README.md
isuftin: Sounds good!

Docker/README.md
zmoore-usgs: Yeah, I wasn't sure if it was for Docker Machine or something. I didn't have to use it myself when running locally. I'll remove it

coastal-hazards-liquibase/Dockerfile
```diff
@@ -13,13 +12,29 @@ ENV POSTGRES_USER=postgres
...
 
-RUN if [ "${doi_network}" = true ]; then \
-		mkdir -p /usr/lib/ssl/certs/ && \
-		/usr/bin/wget -O /usr/lib/ssl/certs/DOIRootCA.crt http://sslhelp.doi.net/docs/DOIRootCA2.cer && \
-		ln -sf /usr/lib/ssl/certs/DOIRootCA.crt /usr/lib/ssl/certs/`openssl x509 -hash -noout -in /usr/lib/ssl/certs/DOIRootCA.crt`.0 && \
-		echo "\\n\\nca-certificate = /usr/lib/ssl/certs/DOIRootCA.crt" >> /etc/wgetrc; \
+# Add Alpine 3.6 repository to get access to plpython
+RUN echo "@v3.6 http://dl-cdn.alpinelinux.org/alpine/v3.6/main/" >> /etc/apk/repositories
+
+RUN set -x \
```
isuftin: Why do you `set -x` here?

zmoore-usgs: Not sure, it was being done before (see line 47) so I left it in

Docker/keycloak/cch_local_development_realm.json
```diff
@@ -0,0 +1,1800 @@
...
+                "view-users",
+                "manage-events",
+                "query-groups",
+                "create-client",
+                "view-events"
+              ]
+            }
+          },
+          "clientRole": true,
+          "containerId": "df79cade-ed8b-4f36-ae19-1ee6bec0d52c",
```
isuftin: This value is repeated throughout. Could you not put this into `Docker/keycloak/keycloak_tomcat_config.json` ?

isuftin: Actually never mind. I thought you were doing variable subs from that file into this in `${}` blocks .. ?

zmoore-usgs: This file isn't meant to be created or edited by hand - It's a generated realm export for Keycloak to re-import the realm when it starts up. I setup the realm in Keycloak once and then used its built-in export function to create this.

`keycloak_tomcat_config.json` is different - It's the file that's loaded by the portal at runtime to connect to the realm that Keycloak loads from the other JSON file

zmoore-usgs: `keycloak_tomcat_config.json` Is also auto-generated by Keycloak itself and isn't meant to be created or edited by hand.

coastal-hazards-liquibase/Dockerfile
zmoore-usgs: I removed it

Docker/README.md
zmoore-usgs: Removed

Docker/keycloak/cch_local_development_realm.json
zmoore-usgs: Yeah, whatever subs are defined here are all created and handled within Keycloak itself

isuftin: I wonder if this is something that can be done during the image built (RUN command) or container start instead of including the auto-generated file in the repo?

coastal-hazards-portal/Dockerfile
```diff
@@ -1,37 +1,41 @@
...
 COPY docker/key-store.jks /usr/local/tomcat/ssl/key-store.jks
 
 COPY docker/key-store.jks /usr/local/tomcat/ssl/trust-store.jks
 
-RUN wget -O /usr/local/tomcat/webapps/coastal-hazards-portal.war "https://cida.usgs.gov/maven/service/local/repositories/cida-public-releases/content/gov/usgs/cida/coastalhazards/coastal-hazards-portal/${PORTAL_VERSION}/coastal-hazards-portal-${PORTAL_VERSION}.war"
+RUN curl -o /usr/local/tomcat/webapps/coastal-hazards-portal.war "https://cida.usgs.gov/maven/service/local/repositories/cida-public-releases/content/gov/usgs/cida/coastalhazards/coastal-hazards-portal/${PORTAL_VERSION}/coastal-hazards-portal-${PORTAL_VERSION}.war"
 
 RUN curl -o /usr/local/tomcat/lib/postgresql.jar https://jdbc.postgresql.org/download/postgresql-$POSTGRES_JDBC_VERSION.jar
 
+RUN curl -o /usr/local/tomcat/keycloak.tar.gz https://downloads.jboss.org/keycloak/$KEYCLOAK_ADAPTER_VERSION/adapters/keycloak-oidc/keycloak-tomcat8-adapter-dist-$KEYCLOAK_ADAPTER_VERSION.tar.gz
```
isuftin: Consider doing these three steps in one RUN command to minimize the amount of space used by by layers inn the final docker image

coastal-hazards-liquibase/Dockerfile.local
```diff
@@ -13,13 +12,28 @@ ENV POSTGRES_USER=postgres
...
+    		curl \
+    		openjdk8 && \
+	rm -rf /var/cache/apk/*;
+
+RUN if getent ahosts sslhelp.doi.net > /dev/null 2>&1; then \
+		mkdir -p /etc/ssl/certs && \
+		wget -O /usr/local/share/ca-certificates/DOIRootCA.crt "http://sslhelp.doi.net/docs/DOIRootCA2.cer"; \
 	fi
 
+RUN update-ca-certificates
```
isuftin: This command (in all images that do this) can probably just be moved into the if statement and run after `wget`. After all, there's no need to run this if we're not importing the DOI cert

coastal-hazards-portal/docker/context.xml
```diff
@@ -32,16 +36,6 @@
 	<Environment name="coastal-hazards.portal.geoserver.endpoint" type="java.lang.String" value="${cch_portal_geoserver_endpoint}"/>
 	<Environment name="coastal-hazards.n52.endpoint" type="java.lang.String" value="${cch_n52_endpoint}"/>
 	<Environment name="coastal-hazards.files.upload.max-size" type="java.lang.String" value="${cch_upload_max_size}"/>
-	<!-- TODO: Is this used? -->
```
isuftin: Decided this is no longer used?

Docker/keycloak/cch_local_development_realm.json
zmoore-usgs: I thought about that, but in the case of not running Keycloak in Docker (or trying to hit the real Keycloak) these files will have to change (and the big one wouldn't exist at all) so bundling them in the image probably wouldn't work. I'm not building the Keycloak image, just using it directly from: https://hub.docker.com/r/jboss/keycloak/ That image supports loading a realm JSON export at startup which is what the 1800 line JSON file is for.


For the portal itself the Tomcat Keycloak Adapter has to load the keycloak configuration from a physical file, and the contents depend on the configuration of Keycloak itself (you grab that JSON from the Keycloak UI once you configure the client) so I'm afraid it'd be too dynamic to do bake it into the image build. The only other way to get the config JSON is to pull it from the Keycloak API, but then you'd need the Keycloak instance you're going to use to be running when you build the portal image.

coastal-hazards-portal/docker/context.xml
zmoore-usgs: Yep

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -137,10 +130,10 @@ public Response searchItems(
 	 * @param request passed through context of request
 	 * @return
 	 */
-	@RolesAllowed({CoastalHazardsTokenBasedSecurityFilter.CCH_ADMIN_ROLE})
 	@POST
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Produces(MediaType.APPLICATION_JSON)
+	@ConfiguredRolesAllowed("coastal-hazards.portal.auth.admin.role")
```
isuftin: You're repeating this string throughout the project. You might consider adding this as a public static String in `ConfiguredRolesAllowedDynamicFeature`

zmoore-usgs: Changed

coastal-hazards-liquibase/Dockerfile.local
zmoore-usgs: Changed

coastal-hazards-portal/Dockerfile
zmoore-usgs: Changed

Docker/keycloak/cch_local_development_realm.json
isuftin: Understood. That's unfortunate

coastal-hazards-portal/src/main/webapp/js/application/publish/OnReady.js
```diff
@@ -28,5 +28,5 @@ $(document).ready(function () {
 		item: CCH.itemid || ''
 	});
 
-	CCH.Auth.checkAuthStatus().then(CCH.publish.init);
+	CCH.publish.init();
```
isuftin: So I am assuming that with this change that the user will need to be auth'd before coming into the application at this route? And this is then handled at the Servlet layer before being allowed in here?

isuftin: Based on the above web.xml, it doesn't seem that way...

coastal-hazards-portal/src/main/webapp/META-INF/context.xml
```diff
@@ -1,2 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Context antiJARLocking="true" path="/coastal-hazards-portal"/>
+<Context antiJARLocking="true" path="/coastal-hazards-portal">
```
isuftin: I don't think `antiJARLocking` is part of Tomcat anymore

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/util/ogc/WFSService.java
```diff
@@ -85,7 +85,7 @@ public boolean checkValidity() {
         boolean isValid = false;
         try {
             isValid = !(StringUtils.isBlank(this.getEndpoint()) || StringUtils.isBlank(this.getTypeName()));
-            URL testUrl = new URL(this.getEndpoint());
+            new URL(this.getEndpoint()).toString();
         } catch (MalformedURLException ex) {
             isValid = false;
```
isuftin: You probably don't need that here since this variable starts out as false anymore. You can essentially catch and do nothing here or log it trace or debug

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/util/ogc/WFSService.java
```diff
@@ -85,7 +85,7 @@ public boolean checkValidity() {
         boolean isValid = false;
         try {
             isValid = !(StringUtils.isBlank(this.getEndpoint()) || StringUtils.isBlank(this.getTypeName()));
-            URL testUrl = new URL(this.getEndpoint());
+            new URL(this.getEndpoint()).toString();
```
isuftin: Why add the toString() here if you don't assign this to anything?

isuftin: You can also:

```java
    public boolean checkValidity() {
        try {
            isValid = !(StringUtils.isBlank(this.getEndpoint()) || StringUtils.isBlank(this.getTypeName()));
            new URL(this.getEndpoint()).toString();
        } catch (MalformedURLException ex) {
            return false;
        }
        return true;
    }
```

isuftin: or...

```    public boolean checkValidity() {
        try {
            !(StringUtils.isBlank(this.getEndpoint()) || StringUtils.isBlank(this.getTypeName()));
            new URL(this.getEndpoint()).toString();
            return true;
        } catch (MalformedURLException ex) {
            return false;
        }
    }
```

coastal-hazards-portal/src/main/webapp/js/application/publish/OnReady.js
zmoore-usgs: The `web.xml` is configuring the Keycloak Tomcat Adapter authentication/authorization. It's the first filter in the auth chain, and then the `ConfigurableRolesAllowedFeature` is the last filter in the auth chain. Since you can't supply values to the `web.xml` at runtime via configuration you have to hard-code the allowed roles in the `web.xml`. In order to get around that and allow runtime configurable roles I have the `web.xml` auth (which is handled by the Keycloak Tomcat Adapter) allow any authenticated user to continue down the chain (which is what the wildcard auth-constraint does). Then they hit the `ConfigurableRolesAllowedFeature` filter in the chain which compares their roles to the configured list and denies them if they don't have the correct role. So at this point if they don't have the configured admin role they'll be rejected at the route by the `ConfigurableRolesAllowedFeature` filter, before this JS file would be fetched.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/util/ogc/WFSService.java
zmoore-usgs: Need to cast the URL to a boolean because it complains about an invalid comparison between `boolean` and `URL`. So I did:  `(!(StringUtils.isBlank(this.getEndpoint()) || StringUtils.isBlank(this.getTypeName()))) && new URL(this.getEndpoint()) != null`

isuftin: Sorry... I think you can do the following:

```java
    public boolean checkValidity() {
        try {
            new URL(this.getEndpoint()); // This wouldn't work if this.getEndpoint() were blank         
            return StringUtils.isNotBlank(this.getTypeName());
        } catch (MalformedURLException ex) {
            isValid = false;
        }
    }
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/HttpUtil.java
```diff
@@ -51,13 +50,11 @@ public static URI convertUriToHttps(URI uri){
 	public static String fetchDataFromUri(String endpoint) {
 		String data = null;
 
-		try {
+		try(DefaultHttpClient client = new DefaultHttpClient()) {
```
isuftin: DefaultHttpClient has been deprecated. You probably want:

```java
try(CloseableHttpClient client = HttpClientBuilder.create().build()) {
 ...
}
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TreeResource.java
```diff
@@ -349,4 +348,8 @@ private boolean deleteItems(JsonArray toDelete, boolean deleteChildren) {
 
 		return returnStatus;
 	}
-}
+
+	private Boolean stringIsNotBlank(String str) {
```
isuftin: Why reinvent StringUtils.isNotBlank()?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
```diff
@@ -0,0 +1,104 @@
+package gov.usgs.cida.coastalhazards.rest.security;
```
isuftin: Did you copy this from another project?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TreeResource.java
zmoore-usgs: I had removed the dependency that included it and just wrote my own replacement, but then I ended up getting it back with some of the Keycloak stuff, so I'll revert this.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
```diff
@@ -0,0 +1,104 @@
...
+			this.denyAll = true;
+			this.rolesAllowed = null;
+		}
+
+		ConfiguredRolesAllowedPastLoginFilter(String[] rolesAllowed) {
+			this.denyAll = false;
+			this.rolesAllowed = (rolesAllowed != null) ? rolesAllowed : new String[] {};
+		}
+
+		@Override
```
isuftin: Formatting for this function is weird

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
```diff
@@ -0,0 +1,104 @@
...
+						return;
+					}
+				}
+            }
+            
+            throw new ForbiddenException();
+		}
+	}
+
+	protected String[] loadRoles(String property) {
```
isuftin: Formatting here is weird too

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
zmoore-usgs: It's just renamed from this project, but I think I rewrote enough of it that it's treating it as a new file. It started out as https://github.com/USGS-CIDA/coastal-hazards/pull/1678/files/72854a50a6852b5a90e639ae6121574518d717ab#diff-375b0f157864e2cb36afeeaa6489aadd but then I changed it a fair bit

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
```diff
@@ -0,0 +1,104 @@
...
+        DynamicReadOnlyProperties props = JNDISingleton.getInstance();
+        
+        // Default value in getProperty doesn't assign to empty values, so we have to do a maunal check after
+        String adminRole = props.getProperty(property, "");
+        
+        if(adminRole == null || adminRole.trim().isEmpty()) {
```
isuftin: `if (StringUtils.isBlank(adminRole)) {}`

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
zmoore-usgs: (src/main/java/gov/usgs/cida/coastalhazards/rest/security/DynamicRolesLoginRedirectFeature.java)

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/security/ConfiguredRolesAllowedDynamicFeature.java
```diff
@@ -0,0 +1,104 @@
...
+		private final String[] rolesAllowed;
+
+		ConfiguredRolesAllowedPastLoginFilter() {
+			this.denyAll = true;
+			this.rolesAllowed = null;
+		}
+
+		ConfiguredRolesAllowedPastLoginFilter(String[] rolesAllowed) {
+			this.denyAll = false;
+			this.rolesAllowed = (rolesAllowed != null) ? rolesAllowed : new String[] {};
```
isuftin: if rolesAllowed is null, doesn't that make denyAll effectively true?

zmoore-usgs: Yes, but that's how the original class is implemented and I didn't want to change it significantly. But I think I've changed enough at this point that it's probably fine.

isuftin: 👍

isuftin:
That's all the comments I have for this PR :)

PR #1684, opened 2019-11-14
Change travis to use openjdk

mhines-usgs:
last week, Bill and I noticed travis was failing the builds and it appears as though it's because we were using oraclejdk8, so i switched it to openjdk8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ services:
   - docker
 
 jdk:
-  - oraclejdk8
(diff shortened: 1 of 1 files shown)


isuftin:
@mhines-usgs Looks like the dependencies for building the rserve Docker image are once again b0rk'd

mhines-usgs:
@isuftin  dur, ok. will address that here, too.

PR #1686, opened 2019-12-03
Refactoring of local development Dockerfiles for easier setup via con…

zmoore-usgs:
…figuration mounting and host networking.

Changes:

- Adjusted Dockerfiles so that they mount Tomcat configuration files at runtime instead of backing them in during the image build which makes it faster and easier to update them and helps reduce the abstraction layers for configuration parameters. Now parameter values are just set directly in the tomcat XML files instead of via `setEnv` scripts like before. This does introduce a bit more configuration duplication, but I think it's a bit easier to follow than tracing xml parameters to setEnv parameters to compose.env environment variables like before.

- Adjust the network mode for most containers to "host" so that they can access eachother via "localhost". This simplifies configuration and helps reduce issues where the same configuration URL needs to work inside and outside of the Docker network (such as the portal base URL parameter which GeoServer hits for SLDs and clients using the UI hit for some web requests).

*Note*: As mentioned on Slack using host network mode will likely limit the docker compose configuration to working only on Docker for Linux

This does not include any changes for Tomcat 8.5.


coastal-hazards-geoserver/Dockerfile
```diff
@@ -20,28 +20,18 @@ RUN apk add --update --no-cache \
...
   rm /var/cache/apk/*
 
-RUN mkdir -p /data/coastal-hazards/uploaded-data
-
-RUN mkdir -p /usr/local/tomcat/ssl
-
-COPY docker/key-store.jks /usr/local/tomcat/ssl/key-store.jks
-
-COPY docker/key-store.jks /usr/local/tomcat/ssl/trust-store.jks
+RUN mkdir -p /data/coastal-hazards/uploaded-data && mkdir -p /usr/local/tomcat/ssl
```
isuftin: This can be `RUN mkdir -p /data/coastal-hazards/uploaded-data /usr/local/tomcat/ssl`

mkdir takes a variable amount of directories as the last item(s)

coastal-hazards-geoserver/Dockerfile
```diff
@@ -20,28 +20,18 @@ RUN apk add --update --no-cache \
...
-COPY docker/tomcat-users.xml /usr/local/tomcat/conf/tomcat-users.xml
-
-COPY docker/context.xml /usr/local/tomcat/conf/context.xml
-
-COPY docker/server.xml /usr/local/tomcat/conf/server.xml
-
 RUN chmod +x /usr/local/tomcat/bin/setenv.sh
 
+RUN rm -rf /usr/local/tomcat/webapps/ROOT /usr/local/tomcat/webapps/docs /usr/local/tomcat/webapps/examples
+RUN rm -rf /usr/local/tomcat/conf/tomcat-users.xml /usr/local/tomcat/conf/server.xml /usr/local/tomcat/conf/context.xml
```
isuftin: No need to have the `-r` flag here if you're just deleting files

coastal-hazards-geoserver/docker/context.xml
```diff
@@ -3,5 +3,5 @@
 	<WatchedResource>WEB-INF/web.xml</WatchedResource>
 	<Manager pathname="" />
     <Environment name="gov.usgs.cida.coastalhazards.wps.fetch.and.unzip.process.unzip.base" type="java.lang.String" value="/data/coastal-hazards/uploaded-data" override="false" /> 
-    <Environment name="gov.usgs.cida.coastalhazards.wps.fetch.and.unzip.process.token" type="java.lang.String" value="${cch_fetch_and_unzip_token}" override="false" />
+    <Environment name="gov.usgs.cida.coastalhazards.wps.fetch.and.unzip.process.token" type="java.lang.String" value="I_pf^-qRE~^8Sf#" override="false" />
```
isuftin: ? What's this?

zmoore-usgs: No idea, but it's what `compose.env` was injecting for this param in geoserver, portal, and n52wps

zmoore-usgs: See: https://github.com/USGS-CIDA/coastal-hazards/pull/1686/files#diff-d6c9597eaad89b70101a8263e3284764L42

README.md
```diff
@@ -11,58 +11,68 @@ the science behind coastal change hazards and providing data, tools, and
...
-instructions make the assumption they are executed from the coastal hazards
-project root directory.
+#### TL;DR - Just get it running
 
-#### Setup
-1. Change `EXTERNAL_HOST` to have the value of your host machine's IP 
-(or localhost if running Docker directly).
-    * using docker-machine: `docker-machine ip <machine name>`
-    * Ubuntu 18.04: `hostname -I`
+1. From project root directory `mvn clean package`
```
isuftin: I suppose at some point we should incorporate a Java build container that can perform this instead of doing it on the host

README.md
```diff
@@ -11,58 +11,68 @@ the science behind coastal change hazards and providing data, tools, and
...
+#### Setup
+1. If building from local sources, run `mvn clean package`. There should be a
    total of 3 WAR files created, 1 in each of `coastal-hazards-n52`,
    `coastal-hazards-geoserver`, and `coastal-hazards-portal`.
 
 #### Building and running the docker containers
 
-1. To build the images, execute `docker-compose -f docker-compose.yml -f
-   docker-compose-local.yml build`
+1. To build the images from local sources execute
```
isuftin: Maybe better worded as `To build the images from artifacts built locally execute ...`

README.md
```diff
@@ -99,13 +109,13 @@ This will bring down _all_ of the running CCH services defined in the
...
-An alternative method for only brining down select services is to run `docker
-ps` to find the `container ID` of the service that you'd like to bring down.
-Once you've found the `container ID` run `docker stop <container ID>` to stop
-that service and then, if you'd like to also remove the container, run `docker
-rm <container ID>`.
+An alternative method for only brining down select services is to run 
+`docker ps` to find the container ID of the service that you'd like to bring down.
+Once you've found the container ID run `docker stop <container ID>` to stop
+that service and then, if you'd like to also remove the container, run 
+`docker rm <container ID>`.
```
isuftin: Because we use Docker Compose here, a user can also just run `docker-compose stop <compose service name` (e.g. `docker-compose stop cch_n52_wps`)

coastal-hazards-geoserver/docker/context.xml
isuftin: Weird. I don't remember what that is

isuftin:
Have you tested this locally?

coastal-hazards-geoserver/docker/context.xml
mhines-usgs: i am not sure why it came about but I believe Carl and Suzanne introduced it when we added raster data support to the portal?

isuftin: magicstrings++

mhines-usgs:
I haven't tested it yet, but am planning to this afternoon

isuftin:
I have no issues with this PR if it does work locally. The changes do make sense though I am not a big fan of using host mode networking but understand why it's being done in this specific case.

isuftin:
I did notice in the travis jobs that the clair scanner is not working as expected :( Maybe something to fix later on

README.md
zmoore-usgs: Woah, has that always been possible? I thought I had tried that at one point but couldn't get it to work so I always do it this way. Good to know

README.md
zmoore-usgs: Probably so.. Not going to include that here, but would be a good task to consider for this FY as part of our meetings this week

zmoore-usgs:
I've tested locally and it works for me. I'm trying out Tomcat 8.5 stuff now

README.md
zmoore-usgs: Will include this update in my Tomcat 8.5 PR because I've already committed some Tomcat 8.5 changes locally

README.md
zmoore-usgs: Yeah, probably at some point. Won't include it in this PR but will note that for a potential ticket for this FY

zmoore-usgs:
Going to go ahead and merge this PR and then create a follow-up with a couple of the changes Ivan suggested plus Tomcat 8.5 stuff

PR #1688, opened 2020-01-10
update rserve library versions

mhines-usgs:


mhines-usgs:
@bcondon-wh  -- if you have feedback of course! otherwise, it's not entirely necessary. One other thing you might look at (or I could show you next time we meet in another pull request "PR" would be in line comments on the code itself.

Docker/rserve/Dockerfile
```diff
@@ -12,8 +12,8 @@ RUN apt-get update && \
 	zlib1g-dev=1:1.2.8.dfsg-2+b1 \
 	libssh2-1=1.4.3-4.1+deb8u6 \
 	libpng12-dev=1.2.50-2+deb8u3 \
-	libxml2=2.9.1+dfsg1-5+deb8u7 \
-	libxml2-dev=2.9.1+dfsg1-5+deb8u7 && \
+	libxml2=2.9.1+dfsg1-5+deb8u8 \
```
mhines-usgs: here is an interesting comment about this line of code, or a question, etc.

mhines-usgs: @bcondon-wh  here is what I mean when I say 'in line comment'

PR #1690, opened 2020-02-07
More Local Development Docker Refactoring

zmoore-usgs:
- Removed maven_base image and moved builds for specific portions into each service Dockerfile. Utilized maven's ability to only build specific sub-modules to build only the parts of the project needed for each Dockerfile. 

  - Did this because there were frequent issues cropping up of trying to rebuild a specific service from local sources failing due to maven_base not being accessible. Doing this also means that when changing only a single sub-module of the project we don't have to rebuild everything.

- Reversed the local vs. remote docker builds so that the default is to build from local sources and building from remote artifacts requires injecting a second compose file.

- Rewrote most of the main readme

- Fixed an issue where KeyCloak would sometimes fail to import the CCH Realm because it got triggered too quickly. Replaced the built-in realm loading behavior with a script which loads the realm only after KeyCloak is up and listening.


Docker/README.md
```diff
@@ -38,48 +45,49 @@ difference you should experience if using just Docker is that instead of
...
 
-#### Running The Portal and GeoServer locally with all other services in containers
 ================
 
 Run a command like the following to stand up the relevant containers:
 
-`
+```bash
 docker-compose up cch_keycloak cch_postgres cch_rserve
 cch_n52_wps
```
isuftin: This should be on the same line as the rest

Docker/keycloak/load_cch_realm.sh
```diff
@@ -0,0 +1,5 @@
+# Waits until the keycloak server is fully up and running and then uses the CLI to import the CCH realm
```
isuftin: Throw a hashbang on top of this

Docker/keycloak/start_cch_keycloak.sh
```diff
@@ -0,0 +1,3 @@
+# Executed on JBoss startup, starts the load_cch_realm script in the background
```
isuftin: Add a hashbang. Also, is there a point to this file? Why not just call the file in the first line directly?

It's unclear what calls this?

coastal-hazards-geoserver/Dockerfile
```diff
@@ -1,37 +1,62 @@
...
+		mkdir -p /etc/ssl/certs && \
+		wget -O /usr/local/share/ca-certificates/DOIRootCA.crt "http://sslhelp.doi.net/docs/DOIRootCA2.cer" && \
+		update-ca-certificates; \
+	fi
+
+WORKDIR /source
+
+COPY pom.xml pom.xml
+
+COPY ehcache-shaded/pom.xml ehcache-shaded/pom.xml
```
isuftin: Are you copying all of these in because the parent pom is going to look for those poms?

coastal-hazards-geoserver/Dockerfile
```diff
@@ -1,37 +1,62 @@
...
+RUN mvn -B -pl coastal-hazards-geoserver -am package
+
+FROM tomcat:8.5-jre8-alpine
 
-ARG GEOSERVER_VERSION=1.1.63
 ENV KEY_STORE_PASSWORD=changeit
 
 RUN apk add --update --no-cache \
 	openssl \
-	fontconfig \
```
isuftin: I remember these being necessary for displaying certain fonts to the client. Are you sure we no longer need these?

coastal-hazards-geoserver/Dockerfile
zmoore-usgs: Yeah, unfortunately event with the maven flag to only build specific modules it still needs to scan all of the poms or it fails

coastal-hazards-geoserver/Dockerfile.remote
```diff
@@ -1,37 +1,34 @@
...
 
 RUN apk add --update --no-cache \
 	openssl \
-	fontconfig \
-	ttf-dejavu \
 	curl && \
 	rm -rf /var/lib/apt/lists/* && \
   rm /var/cache/apk/*
 
 RUN if getent ahosts "sslhelp.doi.net" > /dev/null 2>&1; then \
```
isuftin: We seem to copy/paste this in each image. I wonder if it's useful to make a base image for this as part of the build and just copy the certificate bundle from that base image in each of the java-based containers?

Just a thought.

Docker/keycloak/start_cch_keycloak.sh
zmoore-usgs: So the jboss/keycloak docker image supports running custom scripts on startup by executing any script file mounted into a specific directory before starting up JBoss. (See `Running custom scripts on startup` from: https://hub.docker.com/r/jboss/keycloak/)

However, custom scripts that are executed are blocking and prevent JBoss from actually starting up until they exit, so I have this wrapper script which executes the real listener script asynchronously and then just exits so that JBoss can finish starting up.

coastal-hazards-geoserver/Dockerfile.remote
```diff
@@ -1,37 +1,34 @@
...
-		echo "\\n\\nca-certificate = /usr/lib/ssl/certs/DOIRootCA.crt" >> /etc/wgetrc; \
+		mkdir -p /etc/ssl/certs && \
+		wget -O /usr/local/share/ca-certificates/DOIRootCA.crt "http://sslhelp.doi.net/docs/DOIRootCA2.cer" && \
+		update-ca-certificates; \
 	fi
 
-RUN mkdir -p /data/coastal-hazards/uploaded-data /usr/local/tomcat/ssl
+RUN mkdir -p /usr/local/tomcat/ssl
+
+RUN curl -o /usr/local/tomcat/webapps/probe.war "https://repo1.maven.org/maven2/com/github/psi-probe/psi-probe-web/3.3.1/psi-probe-web-3.3.1.war"
```
isuftin: Same idea for a base image that also happens to download this the one time and we copy it from there. Maybe like a "java-bootstrap" image for the downstream java suite of images during the build 🤷‍♂

coastal-hazards-geoserver/Dockerfile.remote
zmoore-usgs: Yeah, I think when we actually start building images for deployment in CHS for this project we'll definitely want to do a base image for CCH Tomcat. But since these are just for local development right now I figured it was okay to just copy-paste it in.

Docker/keycloak/start_cch_keycloak.sh
isuftin: Gotcha. I understand why this is here now. And I suppose it';s so you can launch the script into the background via `&`.

coastal-hazards-n52/Dockerfile
```diff
@@ -1,31 +1,59 @@
...
+COPY coastal-hazards-export/pom.xml coastal-hazards-export/pom.xml
+COPY coastal-hazards-n52/pom.xml coastal-hazards-n52/pom.xml
+COPY coastal-hazards-wps/pom.xml coastal-hazards-wps/pom.xml
+COPY coastal-hazards-liquibase/pom.xml coastal-hazards-liquibase/pom.xml
+COPY coastal-hazards-geoserver/pom.xml coastal-hazards-geoserver/pom.xml
+COPY coastal-hazards-portal/pom.xml coastal-hazards-portal/pom.xml
+
+RUN mvn -B -pl coastal-hazards-n52 -am dependency:go-offline
 
-LABEL maintainer="gs-w_eto_eb_federal_employees@usgs.gov"
```
isuftin: No maintainer label?

coastal-hazards-geoserver/Dockerfile.remote
zmoore-usgs: Yeah, agreed. I think a base image for CCH Tomcat stuff would be very beneficial, but I wanted to allow all of these images to be able to be built locally right now without order mattering. 

If we had a base image added now it'd need to be built first locally and if it got deleted somehow and then someone tried to rebuild one of the images that depended on it without rebuilding it first they'd error out. 

If we create a base image and then get it hosted up in a registry somewhere so that builds can just pull it I think that'd be sweet

coastal-hazards-n52/Dockerfile
zmoore-usgs: 🤷‍♂️ WMA isn't planning to be the maintainer of this for much longer with this transition and these images aren't used by anything other than us so I figured it wasn't too important. Can add it back later if we want to and if we have an email for whatever group will be taking over

coastal-hazards-geoserver/Dockerfile
zmoore-usgs: Hmm, maybe you're right.. I'll add them back just to be safe

isuftin:
I'm kind of confused why change .local to .remote?

zmoore-usgs:
To reverse the build style that requires the longer command.

I think we build from local sources _way_ more often than the remote artifacts, so I thought it made sense for building from local sources to be the default when you just run `docker-compose up` and then have pulling from remote artifacts be the non-default which requires a longer command

Docker/keycloak/start_cch_keycloak.sh
zmoore-usgs: Yep

.travis.yml
```diff
@@ -31,7 +31,7 @@ before_install:
   - git clone https://github.com/docker/docker-bench-security.git
 
 install:
-  - for image in $DOCKER_IMAGE_BUILD; do docker-compose up -d --build $image; done
+  - for image in $DOCKER_IMAGE_BUILD; do docker-compose -f docker-compose.yml -f docker-compose-remote.yml up -d --build $image; done
```
isuftin: Looks like Travis is beginning to fail builds...

https://travis-ci.org/USGS-CIDA/coastal-hazards/jobs/647498676?utm_medium=notification&utm_source=github_status

coastal-hazards-geoserver/Dockerfile.remote
isuftin: I think that's fine

.travis.yml
zmoore-usgs: Yeah, I found the issue with that. Forgot to change some COPY paths for the `.remote` Dockerfiles

isuftin:
Oh. I now understand local vs remote ;)

zmoore-usgs:
Yeah, I dunno if "remote" is the right term, but I couldn't think of anything better 🤷‍♂️

isuftin:
Travis is looking good

isuftin:
I'm good w/ this

zmoore-usgs:
Cool, thanks for taking a look. I'll make a note to create a ticket to look into creating a base image for this stuff when we start making images for actual deployment instead of just local development.

zmoore-usgs:
@bcondon-wh @zxue-usgs Feel free to take a look at this. These are the changes I mentioned at our check-in this morning and on Teams.

mhines-usgs:
checking this out locally now!

Docker/README.md
```diff
@@ -38,48 +45,48 @@ difference you should experience if using just Docker is that instead of
...
+containers should automatically be able to detect when being built on the DOI
 Network and will make the necessary modifications to be able to pull
 dependencies when this is the case.
 
+### Building Everything
 
-#### Building Everything
 ==========
+
 Clone the docker-rserve repo and build it as described in the first paragraph of
```
mhines-usgs: is this a dependency that we'll remove this year? also, is this new or has it been there and it's just not been as noticeable before? just wondering if we're moving off of wma/cida stuff into a more enterprise/less dependent on cida/wma stuff if this is a good thing to keep or if we should consider the need to remove

isuftin: Are you asking whether we're removing rserve this year?

zmoore-usgs: The dependency on Rserve? I don't think we have plans to remove that this year as far as I remember. I think it's still used for both thumbnail generation and N52 WPS (though what that uses Rserve for I'm not sure)

You don't need to clone the repo to run CCH locally, but if you want to build the Rserve container from local sources instead of pulling a specific version of it then you have to.

mhines-usgs: Hm ok, from the readme, it reads to me like step 1 is to go clone and build this and then do the other steps https://github.com/USGS-CIDA/coastal-hazards/blob/35d8da519718385a273552474f3c8281a872e02d/Docker/README.md#building-everything

mhines-usgs: I meant remove the dependency on our specific WMA created docker-rserve stuff, not necessarily remove rserve from the portal

zmoore-usgs: * Discussed this over on Teams and sorted it out. The readme in the Docker subdir was slightly out of date and WMA's docker-rserve image is no longer a dependency for this project.

zmoore-usgs: I made a commit cleaning up the docker readme a bit, but it's over in this PR because I forgot to switch branches 😅

https://github.com/USGS-CIDA/coastal-hazards/pull/1691

PR #1694, opened 2020-02-11
Fixed a bug where saving child items of a storm...

zmoore-usgs:
... aggregation would cause an error to be thrown due to an invalid attempt to delete the NHC Track item.

### DIFF
--- a/coastal-hazards-liquibase/pom.xml
+++ b/coastal-hazards-liquibase/pom.xml
@@ -27,7 +27,7 @@
 		<pluginRepository>
 			<id>cida-public</id>
 			<name>CIDA plugin repository</name>
-			<url>http://cida.usgs.gov/maven/cida-public/</url>
+			<url>https://cida.usgs.gov/maven/cida-public/</url>
 		</pluginRepository>
 	</pluginRepositories>
     
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
@@ -210,23 +210,22 @@ public class ItemResource {
 				mergedId = itemManager.merge(mergedItem);
 			}
 
+			if (null != mergedId) {
+				response = Response.ok().build();
+			} else {
+				throw new BadRequestException();
+			}
+
 			//Delete the storm track item once the storm item has been successfully saved
-			if (trackId != null && mergedId != null) {
+			if (trackId != null) {
 				itemManager.delete(trackId, true);
 			}
-		}
-
-		if (null != mergedId) {
-			response = Response.ok().build();
-		} else {
-			throw new BadRequestException();
 		}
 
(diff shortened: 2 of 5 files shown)


coastal-hazards-liquibase/pom.xml
```diff
@@ -27,7 +27,7 @@
 		<pluginRepository>
 			<id>cida-public</id>
 			<name>CIDA plugin repository</name>
-			<url>http://cida.usgs.gov/maven/cida-public/</url>
+			<url>https://cida.usgs.gov/maven/cida-public/</url>
```
isuftin: Don't want to switch everything else to https while you're in here?

isuftin: Oops. Didn't scroll

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -181,56 +182,54 @@ public Response postItem(String content, @Context HttpServletRequest request) {
...
 			}
-			try (StatusManager statusMan = new StatusManager(); ThumbnailManager thumbMan = new ThumbnailManager()) {
-				Status status = new Status();
-				status.setStatusName(Status.StatusName.ITEM_UPDATE);
-				statusMan.save(status);
+		}
 
-				thumbMan.updateDirtyBits(id);
-			}
+		if (null != mergedId) {
```
isuftin: Would it make sense to move this above line 213? After all, the if block there won't run if mergedId is null and you can throw early out of that larger try block

isuftin: Also, if you do that, on the if block on line 214 you then don't have to double-check for mergedId being null.

zmoore-usgs: Yeah, that makes sense. It seemed kind of funky to explicitly throw something while within a try block, but that's really not an issue.

isuftin: Yeah the JDK made things a little confusing when combining the functionality of try/catch with try-with-resources :/

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ItemResource.java
```diff
@@ -210,23 +210,22 @@ public Response updateItem(@Context HttpServletRequest request, @PathParam("id")
 				mergedId = itemManager.merge(mergedItem);
 			}
 
+			if (null != mergedId) {
```
isuftin: 👍

mhines-usgs:
Pulling this locally to check out!

mhines-usgs:
works great for me locally, let's merge and test on dev.

PR #1698, opened 2020-03-02
Split minify request parameters, fix tweet button

zmoore-usgs:


zmoore-usgs:
^ Going to ignore the codacy on this one since the established pattern in the UI for CCH is to use single quotes.

PR #1699, opened 2020-03-03
(Mostly) fixed tweet buttons. Removed duplicate dependency from POM.

zmoore-usgs:
- The twitter widget API that we use for generating buttons was updated at some point to have the buttons consider the size of the Div that they're rendering into before being created.
- When we created the tweet buttons we did so within hidden modal divs, and this resulted in tweet buttons being created with a width/height of 1px.
- Moving tweet button creation to after the div is created fixed the sizing issue, however there were still times (perhaps 1/10) when the button would render with the width/height of 1px. Not sure why that happens because the div is fully rendered and should be sized correctly. Assume it's some sort of weirdness with the widget API or something. To fix that I added a retry that checks the size of the button after rendering and if it's 1px then it re-creates it.
- If for some reason that still fails (which happens occasionally) then it shows an error message telling the user to reload the modal and try again.
- I also added a loading spinner while the tweet buttons are waiting to be created to illustrate that something should show up eventually.
- On the "front" UI we were also adding an event listener to show an alertify dialog when the user tweets something every time the share button was clicked, which resulted in multiple alerts popping up if the share button was clicked multiple times. I fixed that by moving the listener creation outside of the modal click handler function.

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
@@ -18,7 +18,7 @@ CCH.Objects.Back.UI = function (args) {
 	me.isSmall = function () {
 		return $(window).outerWidth() < me.magicResizeNumber;
 	};
-	me.DO_RETRY_TWEET_BUTTON = true;
+	me.TWEET_BUTTON_MAX_TRIES = 2;
 
 	me.init = function (args) {
 		me.$qrImage = $('#qr-code-img');
@@ -415,7 +415,7 @@ CCH.Objects.Back.UI = function (args) {
 		});
 	};
 
-	me.createTweetButton = function() {
+	me.createTweetButton = function(currentTry) {
 		twttr.widgets.createShareButton(
 			$('#modal-share-summary-url-inputbox').val(),
 			$('#twitter-button-span')[0],
@@ -432,14 +432,13 @@ CCH.Objects.Back.UI = function (args) {
 			setTimeout(function() {
 				if($('.twitter-share-button')[0].style.width !== "1px") {
 					$('#twitter-button-load').addClass('hidden');
-				} else if(me.DO_RETRY_TWEET_BUTTON) {
+				} else if(currentTry < me.TWEET_BUTTON_MAX_TRIES) {
 					$('#twitter-button-span').empty();
-					me.createTweetButton();
+					me.createTweetButton(currentTry + 1);
 				} else {
 					$('#twitter-button-load').addClass('hidden');
 					$('#twitter-button-span').html("Failed to load tweet button. Please close and try again.")
 				}
-				me.DO_RETRY_TWEET_BUTTON = false;
 			}, 200);
 		});
 	}
@@ -451,8 +450,7 @@ CCH.Objects.Back.UI = function (args) {
 		// Create Tweet button after modal animation (only if it doesn't exist)
 		if($('.twitter-share-button').length == 0) {
 			setTimeout(function() {
-				me.DO_RETRY_TWEET_BUTTON = true;
-				me.createTweetButton();
+				me.createTweetButton(1);
 			}, 300);
 		}
 	});
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
@@ -18,6 +18,7 @@ CCH.Objects.Back.UI = function (args) {
 	me.isSmall = function () {
 		return $(window).outerWidth() < me.magicResizeNumber;
 	};
+	me.DO_RETRY_TWEET_BUTTON = true;
 
 	me.init = function (args) {
 		me.$qrImage = $('#qr-code-img');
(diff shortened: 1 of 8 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
```diff
@@ -428,6 +415,47 @@ CCH.Objects.Back.UI = function (args) {
...
+			}, 200);
+		});
+	}
+
+	$('#modal-sharing-view').on('shown.bs.modal', function() {
+		// Highlight the entire input box
+		$('#modal-share-summary-url-inputbox').select();
+
+		// Create Tweet button after modal animation
+		if(me.DO_RETRY_TWEET_BUTTON) {
```
isuftin: Can this ever be false at this point in the loading process?

coastal-hazards-portal/src/main/webapp/js/cch/objects/front/UI.js
```diff
@@ -553,6 +537,35 @@ CCH.Objects.Front.UI = function (args) {
 		}
 	};
 
+	me.createTweetButton = function(retries) {
```
isuftin: I forget if there's a file that provides functionality to both the back and the front, but I wonder if this block of code that's repeated twice could go to something like that

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
zmoore-usgs: Yeah, this runs every time the "share" button is clicked and `me.DO_RETRY_TWEET_BUTTON` persists between the modal being opened/closed. If the button completely failed to generate or successfully generated we don't want this to run again or it'll create a 2nd button.

This is slightly different from what's done on the "front" because every time you click the share button it might be different contents being shared, where as this button is always for a specific item (whatever one you navigated to).

I could maybe clarify this a bit more

coastal-hazards-portal/src/main/webapp/less/back/back.less
```diff
@@ -282,6 +282,11 @@ html, body {
     right:90px;
 }
 
+.tweet-button-loading {
+    max-width: 20px;
+	margin-right: 20px;
```
isuftin: Tab this back once (or remove two spaces)

coastal-hazards-portal/src/main/webapp/js/cch/objects/front/UI.js
zmoore-usgs: I did look around but couldn't find much. There are also slight differences between the two

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
isuftin: I think it makes sense. It's just in the context of a pull request, by itself it seemed strange

zmoore-usgs: Re-reading what I just wrote I think we actually _do_ what this to run again if the button completely failed to generate the first time, so I'm going to tweak this.

coastal-hazards-portal/src/main/webapp/less/front/application.less
```diff
@@ -199,6 +199,11 @@ html, body {
     bottom: 10px;
 }
 
+.tweet-button-loading {
+    max-width: 20px;
+	margin-right: 20px;
```
isuftin: tabbing

mhines-usgs:
building this locally to check out...

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
isuftin: Perhaps set a max amount of retries and loop through, incrementing a counter.

zmoore-usgs:
@mhines-usgs Going to make a few tweaks, but overall it should be good to test locally

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
zmoore-usgs: I was originally going to do that, but decided I only really wanted a single retry so I went with the boolean instead of a counter. But a counter probably makes more sense from a readability perspective so I can do that

isuftin: Yeah a boolean flip in a retry is always a little weird to read. Even if your counter maxes out at 1, it might be easier to take in

coastal-hazards-portal/src/main/webapp/js/cch/objects/back/UI.js
```diff
@@ -432,14 +432,13 @@ CCH.Objects.Back.UI = function (args) {
 			setTimeout(function() {
 				if($('.twitter-share-button')[0].style.width !== "1px") {
 					$('#twitter-button-load').addClass('hidden');
-				} else if(me.DO_RETRY_TWEET_BUTTON) {
+				} else if(currentTry < me.TWEET_BUTTON_MAX_TRIES) {
 					$('#twitter-button-span').empty();
-					me.createTweetButton();
+					me.createTweetButton(currentTry + 1);
```
isuftin: 👍

isuftin:
Looking good

zmoore-usgs:
@mhines-usgs User-facing behavior is pretty much identical to what it was when you commented so shouldn't need to re-pull but feel free to if you want to.

mhines-usgs:
Thanks, in a meeting but will wrap up poking it right after we finish.

PR #1701, opened 2020-03-06
added UVVR specific rounding in popup

zxue-usgs:
Added UVVR specific rounding to 3 digits while maintaining the 1 decimal place for the rest in popup within cch-portal

### DIFF
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
@@ -366,8 +366,11 @@ return {
 							} else {
 								if (Number.isInteger(displayPoints[i])) {
 									$valueContainer.append(displayPoints[i].toFixed(0));
-								}
-								else {
+								} 
+                                                                //case specific rounding for UVVR
+                                                                else if (attr === 'UVVR') {
+                                                                        $valueContainer.append(displayPoints[i].toFixed(3));
+                                                                } else {
 									$valueContainer.append(displayPoints[i].toFixed(1));
 								}
 							}
--- a/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
+++ b/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
@@ -366,9 +366,8 @@ return {
 							} else {
 								if (Number.isInteger(displayPoints[i])) {
 									$valueContainer.append(displayPoints[i].toFixed(0));
-								} 
-                                                                //case specific rounding for UVVR
-                                                                else if (attr === 'UVVR') {
+								} else if (attr === 'UVVR') { 
+									//case specific rounding for UVVR
                                                                         $valueContainer.append(displayPoints[i].toFixed(3));
                                                                 } else {
 									$valueContainer.append(displayPoints[i].toFixed(1));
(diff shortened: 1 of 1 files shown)


coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -366,8 +366,11 @@ return {
 							} else {
 								if (Number.isInteger(displayPoints[i])) {
 									$valueContainer.append(displayPoints[i].toFixed(0));
-								}
-								else {
+								} 
+                                                                //case specific rounding for UVVR
```
zmoore-usgs: Nitpick: Indent looks a little off here (looks like a tabs vs. spaces thing)

zxue-usgs: image
it looks ok on netbeans

isuftin: ...weird

zmoore-usgs: Yeah, a lot of the time IDEs will have different width rules for tabs which will cause these sorts of issues. It's a very common thing. GitHub prints tabs as actual tabs so on here is where you can see the actual differences.

If you highlight row 369 in the diff here on GitHub you'll see it uses tab characters vs. line 370 which uses spaces. We try to keep the character used consistent (although we do an admittedly poor job of that a lot of the time because it can be hard to tell within an IDE) within a file so that indentation is consistent in different IDEs.

coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js
```diff
@@ -554,4 +553,4 @@ return {
 	},
 	CLASS_NAME: "OpenLayers.Control.WMSGetFeatureInfo"
 }
-}()));
\ No newline at end of file
+}()));
```
zxue-usgs: what's this bit at the end of the code?

zmoore-usgs: It's finishing out all of the closures https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures and function calls that this is nested within.

The outermost `)` is for closing out the call to `OpenLayers.Class(` which the rest of this is nested within on Line 6
The second one is for the wrapper started on Line 7
The `()` set is to have this function called immediately
The `}` is for closing out the function defined on Line 7
The `}` on Line 555 is for closing out the return block of that function, which is started on Line 61

It's very deeply nested and definitely quite confusing to read, but essentially:

We're building an object to be passed into `OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, (<Object>));`

We get the Object as the return value of the function that we define on Line 7 and then execute on Line 556

So it boils down to: 
```javascript
OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, (function() { 
    .... <Run some stuff>
    return: {
        <Object>
    }
}()));
```

Which means: Create and immediate run `function` and pass the returned object into `OpenLayers.Class(OpenLayers.Control.WMSGetFeatureInfo, (<Object>))`

zxue-usgs: I didn't think I made changes to that, why is it showing up as an addition?

zxue-usgs: Oh nevermind, it just changed it.

zmoore-usgs: Yeah looks like the change went from there not being a new line at the end of the file to there being a new line at the end of the file. No biggie, that's totally fine

zxue-usgs: Ah that's what happened, I missed that, thanks!

PR #1702, opened 2020-03-06
Integration tests

zmoore-usgs:


isuftin:
I feel like at some point we did have Travis set up to perform the jmeter integration tests, did we not? Maybe for another project?

zmoore-usgs:
That was for MLR I believe. I thought about it for this, but since we're moving the code repo off here soon anyways and won't be able to use Travis from GitLab I figured it wasn't worth it

isuftin:
Yeah - https://github.com/USGS-CIDA/MLR-Integration-Testing

docker-compose-it.yml
```diff
@@ -0,0 +1,18 @@
+---
+version: '3.5'
```
isuftin: I think 3.7 is the latest version available

docker-compose-it.yml
```diff
@@ -0,0 +1,18 @@
...
+
+services:
+  jmeter:
+    image: justb4/jmeter
+    network_mode: "host"
```
isuftin: I don't think you need quotes here

coastal-hazards-integration-testing/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+### 1. Running Via Docker Compose
+
+To run the integration tests via Docker Compose you should first bring up the full CCH stack via:
+
+`docker-compose up`
+
+Once the stack is fully up and running the integration tests can be run via:
+
+`docker-compose -f docker-compose.yml -f docker-compose-it.yml up jmeter`
```
isuftin: Do you actually need to re-run the docker-compose.yml here?

zmoore-usgs: Not necessarily, but if you don't include the other compose file Docker complains that there it found "orphan containers from this project". It doesn't seem to terminate them, but that could change?

coastal-hazards-portal/docker/server.xml
```diff
@@ -59,6 +59,7 @@
            port="8443" maxThreads="150"
            scheme="https" secure="true" SSLEnabled="true"
            keystoreFile="/usr/local/tomcat/ssl/key-store.jks" keystorePass="changeit"
+           maxSavePostSize="-1"
```
isuftin: Any chance of me causing a buffer overflow here by POSTing gigabytes?

cleanIntegrationTestResults.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+sudo rm -rf ./coastal-hazards-integration-testing/test_results/report
```
isuftin: `sudo rm -rf ./coastal-hazards-integration-testing/test_results/*` ?

coastal-hazards-portal/docker/server.xml
zmoore-usgs: To your own local version yeah, but since we don't use Docker for the real deployments (at least not yet) there's no risk beyond bringing down your own local portal.

cleanIntegrationTestResults.sh
zmoore-usgs: Need to keep the `.keep_me` file so that the folder is in the repo (if the folder doesn't exist compose throws an error when trying to mount it)

docker-compose-it.yml
```diff
@@ -0,0 +1,18 @@
+---
+version: '3.7'
+
+services:
+  jmeter:
+    image: justb4/jmeter
```
isuftin: I see your tests are written in JMeter 5.1.1 and the image you're using happens to use that version but you might consider locking that down by using `justb4/jmeter:5.1.1`

coastal-hazards-portal/docker/server.xml
isuftin: Yep good point. Forgot this was for local only

isuftin: But also if this is not the setting used in not-local, why do this here?

coastal-hazards-integration-testing/README.md
isuftin: Then are you doing this in two parts (docker-compose up and then docker-compose up again) because you need to wait for services to come up?

Take note that unless you daemonize `docker-compose up` via `docker-compose up -d`, your user is now stuck tailing the docker-compose logs and can;t exit from that. If you can get this going by just running the single command `docker-compose -f docker-compose.yml -f docker-compose-it.yml up jmeter` that would work best but if you do have timing issues then you will need to document to add a `-d` flag and then tail the logs via `docker-compose logs -f ...` or open a new shell for the subsequent commands

coastal-hazards-portal/docker/server.xml
zmoore-usgs: It's a quirk with how JMeter executes POST requests that require auth. When executing POSTs from the browser it checks auth requirements and then reads the file input stream. For some reason the HTTP request implementation used by JMeter tries to push the input stream first and then do auth evaluation. This setting is the max POST size that can be stored while Tomcat is verifying authentication and it defaults to 4096 bytes.

I repeated the same request in the browser via an HTTP request tool and in JMeter and JMeter was getting client-side connection exceptions instead of 403's until I added this while the browser was getting 403s. It's weird. 🤷‍♂️

zmoore-usgs: https://stackoverflow.com/questions/57871348/buffer-overflow-error-when-calling-request-authenticate-on-tomcat

cleanIntegrationTestResults.sh
isuftin:
```
sudo rm -rf ./coastal-hazards-integration-testing/test_results/*
sudo touch ./coastal-hazards-integration-testing/test_results/.keep_me
```

or 

`sudo rm -rf ./coastal-hazards-integration-testing/test_results/!(keep_me.txt)`

:D

coastal-hazards-integration-testing/README.md
zmoore-usgs: Yeah, that's the idea and yeah the reason for separate commands is that JMeter will fail if the portal isn't up when it starts running. 

I personally run them in two consoles both in non-daemon mode so I can watch the portal logs from `docker-compose up` and the JMeter logs, but I can note that in the readme and note that daemonized mode can also be used.

cleanIntegrationTestResults.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+sudo rm -rf ./coastal-hazards-integration-testing/test_results/*
+touch ./coastal-hazards-integration-testing/test_results/.keep_me
```
isuftin: So why do you need to use sudo in the first step and not in the second?

zmoore-usgs: The test_results directory itself is owned by the user who clone the repo and thus the `.keep_me` file is/should be as well.

The files that are created in this directory by the containerized JMeter are owned by whatever user UID/GID is used within the container and, at least on my VM, it's not the same as my user UID/GID so I have to use sudo to remove them.

coastal-hazards-integration-testing/README.md
isuftin: The timing issue is unfortunate. If you were building your own jmeter image, you could of course always use something like [wait-for](https://github.com/eficode/wait-for) like we do with [MLR](https://github.com/USGS-CIDA/MLR-Integration-Testing/blob/master/docker-compose-services.yml#L82) :(

zmoore-usgs: Yeah, I considered doing that but figured since for now these are just being run locally it wasn't a huge deal. If/when we have cloud infrastructure with woods hole that we can run these on automatically (or if that PR ever gets merged into GitLab to allow running Docker containers within their Dockerized runners) then it'd probably make sense to do that or have a script that runs outside of the tests that does that and then launches the tests

zmoore-usgs: It also might be possible to do that in the JMeter plan itself with setup thread groups

isuftin: 👍

isuftin:
Looking good!

PR #1709, opened 2020-07-17
UVVR Raster

zxue-usgs:
- Added UVVR Raster functionality based on AE raster files.

-Tweaked legend and layerIdentify .js files including 3-decimal rounding and continuous legend display  to match with polygon uvvr

-Fixed polygon uvvr legend display bug where wrong color was displayed


zxue-usgs:
I'm still running into weird formatting and indentation things. I think it might be something with how netbeans handles tabs and indents? I can fix it manually here.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDConfig.java
```diff
@@ -53,23 +53,23 @@
...
-        this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, SLDGenerator.LEGEND_TYPE_DEFAULT);
-    }
-    public SLDConfig(String jspPath, String units, String style, int strokeWidth, float strokeOpacity, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins, LegendType legendType) {
-        this.jspPath = jspPath;
+        public SLDConfig(String jspPath, String units, String style, int strokeWidth, float strokeOpacity, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins) {
+                this(jspPath, units, style, strokeWidth, SLDGenerator.STROKE_OPACITY_DEFAULT, attrs, thresholds, colors, bins, SLDGenerator.LEGEND_TYPE_DEFAULT);
+        }
+        public SLDConfig(String jspPath, String units, String style, int strokeWidth, float strokeOpacity, String[] attrs, float[] thresholds, String[] colors, List<Map<String,Object>> bins, LegendType legendType) {
+                this.jspPath = jspPath;
```
zxue-usgs, line 62: I tried to fix the indent issues in netbeans but I guess it doesn't actually fix it.

zxue-usgs:
1.Codacy review failed because I named a class RasterUVVR. Should I change it to RasterUvvr so that it's happy?
2.It also complained about the encoding in the new jsp file. I copied it from another raster jsp so the encoding should be the same, right?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/RasterUVVR.java
```diff
@@ -0,0 +1,44 @@
...
+        protected static final List<Map<String, Object>> bins;
+        protected static final float[] thresholds = {0f, 0.025f, 0.050f, 0.075f, 0.100f, 0.200f, 0.300f, 0.400f, 0.500f, 0.75f, 1.0f, 1.25f, 1.5f, 1.75f, 2.0f, 999999999f}; //used to evaluate what color the pixel should receive. (gray_index) 
+        protected static final String[] colors = {"#2B41FF", "#2B41FF", "#386DFF", "#3B9DFF", "#30CFFF", "#00FFFF", "#70FFD2", "#A1FFA4", "#C7FF78", "#E7FF4A", "#FFFF00", "#FFD500", "#FFA600", "#FF7B00", "#FF4D00", "#FF0000"};
+        protected static final float[] range = {-1f, 0f, 0.025f, 0.050f, 0.075f, 0.100f, 0.200f, 0.300f, 0.400f, 0.500f, 0.75f, 1.0f, 1.25f, 1.5f, 1.75f, 2.0f, 999999999f};  //this will be used to create the legend in the map with the ranges etc 
+	protected static final String[] categories;
+        
+	static {
+		categories = new String[range.length-1];
+		for(int i = 1; i < range.length; i++){
+			categories[i-1] = range[i-1] + " to " + range[i];
```
zmoore-usgs, line 23: Since it appears `range` is the same as `thresholds` but with `-1` added to the beginning you could probably do something like this instead to avoid the duplicate definition (so that if some day someone has to change one of the threshold values for some reason they don't have to change it in both thresholds and range):

```
for(int i = 0; i < thresholds.length; i++){
    if(i ==0) {
        // Add lower bound to first threshold for the legend
        categories[i] = "-1 to " + thresholds[i];
    } else {
        categories[i] = thresholds[i-1] + " to " + thresholds[i];
    }
}
````

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/SLDConfig.java
zxue-usgs: @zmoore-usgs it looks like it's a similar issue with netbeans here with tabs and spaces https://github.com/github/VisualStudio/issues/1524
What do you suggest I should do?

zmoore-usgs: Generally what I do is do it somewhat manually. I pick an indent-type for the file (usually whatever has majority at that point, though consistency through the project would be ideal but it's already pretty inconsistent so would be a lot of work to clean up) and then do a search and replace all with that type. So:

1. Going with tabs: Find and replace instances of four space characters with a single tab character (just copy a tab character from the a file and paste into the replace box)

2. Going with spaces: Find and replace instances of the tab character (copy paste into the search box) and replace with four spaces.

Then after that just manually go back and clean up anything that isn't lined up properly. Since it's all using the same character now it should look the same in the editor as it will here

zmoore-usgs:
1. Codacy Review: Yeah, I'd just go ahead and change it since we have the existing Uvvr file that is already lower case like that. That's one of the Codacy rules I don't necessarily agree with and it's a little nit-picky, but just for consistency with that other Uvvr SLD I think changing it makes sense.

2. Yeah, don't worry about that one. Seems like it's a false positive

zxue-usgs:
Codacy review failed again after changing RasterUvvr to PascalCase.

zmoore-usgs:
Huh, not sure what it's mad about now. I'd say you can ignore it and merge!

PR #1713, opened 2020-08-14
Add logging to error page JSP generation.

zmoore-usgs:


zmoore-usgs:
@mhines-usgs Unfortunately despite trying everything I could think of I actually couldn't get the issue to trigger haha. Every error I was able to trigger was properly handled elsewhere in the code and didn't result in this issue showing up (which is good! but kind of funny). So I was only able to test this myself locally by tweaking the `if` statement to just always be true rather than checking for an unhandled error first.

mhines-usgs:
That makes sense, thanks for explaining. I know we have hit it in the past, but I do not remember exactly the situation. Glad you were able to test it out nonetheless!!

PR #1714, opened 2020-08-25
First implementation of FloodMask Polygon dataset support…

zmoore-usgs:
… Slight tweak to error page logging.

Note that this implementation makes some assumptions about the structure of the split shapefiles so may need to be revisited if those assumptions are not attainable or acceptable. I put some details about those assumptions in JIRA and uploaded my modified shapefiles for Florida which include the needed changes for those assumptions to the ticket. I can put the files in Teams as well and I believe Megan copied my explanation of the shapefile assumptions from JIRA into Trello.

Also can definitely make tweaks to the colors and units as desired, I just based the current values off of what I saw in the publication with this data.

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/Attributes.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/Attributes.java
@@ -102,19 +102,21 @@ public class Attributes {
 	public static final String NHC_TRACK_PT_120WLBL = "NHC_TRACK_PT_120WLBL";
 	public static final String NHC_TRACK_PT_72CAT = "NHC_TRACK_PT_72CAT";
 	public static final String NHC_TRACK_PT_120CAT = "NHC_TRACK_PT_120CAT";
-        
-        public static final String UVVR = "UVVR"; //vulnerability index
-	
+
+	public static final String UVVR = "UVVR"; //vulnerability index
+
 	public static final String TCT = "TCT";
 	public static final String TC2 = "TC2";
 	public static final String TC5 = "TC5";
-        
-        //raster attr
-        public static final String GRAY_INDEX = "GRAY_INDEX";
-        public static final String AE = "AE";  //adjusted elevation
-        public static final String PAE = "PAE";  //adjusted elevation probability
-        public static final String CR = "CR";  //coastal response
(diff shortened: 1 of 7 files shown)


mhines-usgs:
This looks ok to me, I think Z is out moving so if you're cool with trying it out on dev, I'd say go for merge. I haven't heard anything from Emily so I'll write her and ask if she has time to meet up for a little while to discuss before Friday..

zmoore-usgs:
Oh right forgot he was out! Alrighty, I'll go for the merge

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/Attributes.java
```diff
@@ -102,19 +102,21 @@
...
 	public static final String TC5 = "TC5";
-        
-        //raster attr
-        public static final String GRAY_INDEX = "GRAY_INDEX";
-        public static final String AE = "AE";  //adjusted elevation
-        public static final String PAE = "PAE";  //adjusted elevation probability
-        public static final String CR = "CR";  //coastal response
-        public static final String UVVR_RASTER = "UVVR_RASTER"; //unvegetated to vegetated ratio, raster
+
+	public static final String FL_MASK_ID = "FL_MASK_ID";
```
zxue-usgs, line 112: For this attribute "FL_MASK_ID", it has to match the table header in the shapefile right? Did we have to change the attribute name in ARC? @zmoore-usgs

zmoore-usgs: Yep, it does have to. I was able to find another application (QGIS) that I could use to edit the shapefile so I actually added that as a new column. It can certainly be renamed or modified, this was just a proposal for one way we could handle it.

Right now the portal doesn't really support dealing with String columns so I added FL_MASK_ID as a numerical column to serve as a "category index" that the portal can key off of (where `1` equates to floodmask without reef and `2` equates to floodmask with reef) which is a pattern that had already been established for "categorical" data like this in the "categorical line" SLD JSP.

I posted more detail in JIRA and I think Megan copied it over to Trello but it's kind of a wall-of-text so if you want to have a quick call about it sometime let me know and I can explain my thinking here.

PR #1715, opened 2020-09-01
Convert to Python 3.8. Replace method for identifying GeoServer…

zmoore-usgs:
… services from CCH items. Update documentation and help.

### DIFF
--- a/util/layer-sweep/README.md
+++ b/util/layer-sweep/README.md
@@ -1,6 +1,6 @@
 # Layer Sweeping
 
-CCH does not always clean up after itself. Sometimes CCH Items that are deleted in the Portal UI are not removed from GeoServer. This directory contains two scripts to help clean up.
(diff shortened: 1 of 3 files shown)


mhines-usgs:
Didn't get to this last evening but am starting with it this morning. :)

mhines-usgs:
Thanks to all of your help, I got it working and was able to identify and delete appropriate layers, and it didn't ignore real time storm forecasts! 

To test I successfully orphaned and deleted an old storm item in the dev UI (first noting one of the geoserver layer iDs for a child item), and then I ran find_dangling_layers to see that geoserverlayer ID, then I modified my output text file to only include that ID, and ran the delete_layers process, and it outputs that it successfully deleted that layer, I rechecked in the dev geoserver UI that I can no longer find that layer.

mhines-usgs (approved):

PR #1717, opened 2020-09-04
Add dependency scanning profile to project parent POM.

zmoore-usgs:
This will run during the `test` phase of a maven build when you have enabled the `security-scan` maven profile. The results are output into the `target/dependency-check` directory of each sub-project in the CCH repo.

Example build commands:

Normal build command with no security scan: `mvn clean package` (This is essentially what the Docker version of the build runs)

Build command with security scan: `mvn -Psecurity-scan clean package`

Note that if you've only been building the portal via Docker you will need to install JDK (the portal uses JDK 8) and maven onto your host machine to run this scan because the version of the build that happens within Docker won't run the scan and although we could configure it to do so it wouldn't be easy to extract the resultant HTML reports out of the container for viewing and the scan would have to re-download the full vulnerability database each time it ran which would significantly add to the build time for the portal during local development.

If you try to run this build with JDK 11 as your default JDK (if your system JAVA_HOME env points to JDK 11 instead of 8) you will see some errors during the build and it won't finish. So you can run it like this to guarantee it uses JDK 8:

`export JAVA_HOME=<path to JDK 8> && mvn -Psecurity-scan clean package`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -554,4 +554,93 @@
 			</plugins>
 		</pluginManagement>
 	</build>
+	<profiles>
+        <profile>
+            <id>security-scan</id>
+            <properties>
+                <owasp.dependency.check.version>5.2.4</owasp.dependency.check.version>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.owasp</groupId>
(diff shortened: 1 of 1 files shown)


zxue-usgs (approved):
Looks straight forward. Thanks for the documentation.

PR #1718, opened 2020-09-11
Implemented shoreline forecasts line and polygons

zxue-usgs:
Added forecast line and polygon sld generation using categorical_line and catagorical_polygon jsp

fixed a few documentations on authors
updated attribute test to include new attributes

### DIFF
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
@@ -23,8 +23,8 @@ public final class ForecastUncy {
 //
 //20-yr line:  R 166, G 97, B 26 #A6601A
 //20-yr polygon:  R 223, G 194, B 125 #DFC27D
-	protected static final String[] attrs = new String[]{FORECASTPE_U};
-	protected static final float[] thresholds = new float[]{1f, 2f};
+	protected static final String[] attrs = new String[]{FORECAST_U};
+	protected static final float[] thresholds = new float[]{10f, 20f};
 	protected static final String[] categories = {"10-year", "20-year"};
 	protected static final String[] colors = {"#80CDC1", "#DFC27D"};
 	protected static final float strokeOpacity = 0.5f;
--- a/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
+++ b/coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
@@ -12,17 +12,6 @@ import java.util.Map;
  */
 public final class ForecastUncy {
 
-//Line thickness:  2.0
-//
-//Polygon outline thickness:  1.0
-//Polygon outline color:  R G B all 110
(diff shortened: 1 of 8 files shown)


coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
```diff
@@ -0,0 +1,51 @@
...
+//
+//Polygon outline thickness:  1.0
+//Polygon outline color:  R G B all 110
+//Polygon transparency:  50%  (in Arc, rt click on shp layer, Properties - Display)
+//
+//10-yr line:  R 1, G 133, B 113 #018571
+//10-yr polygon:  R 128, G 205, B 193 #80CDC1
+//
+//20-yr line:  R 166, G 97, B 26 #A6601A
+//20-yr polygon:  R 223, G 194, B 125 #DFC27D
```
zmoore-usgs, line 25: ^ Feel free to keep the above comments if you want for future reference, though I would recommend changing them to a block comment with a bit of context about what each thing refers to (I.E: The line referring to the other Forescast SLD)

zmoore-usgs (commented):
Overall these changes look good! One thing I'd take a look at though is whether or not the click-to-identify is working as you would expect it to. For my categorical stuff to show up properly I had to add the new attribute I added to the list here: 

https://github.com/USGS-CIDA/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js#L520

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/sld/ForecastUncy.java
zxue-usgs: Thanks for catching that, I deleted these from the other file but forgot about this one.

zxue-usgs:
Codacy is not happy about the "Forecast" and "ForecastUncy" classes

zmoore-usgs:
> Codacy is not happy about the "Forecast" and "ForecastUncy" classes

Yeah, I wouldn't worry about that too much. You could change the `Attributes.*` import to only import the attribute the SLD is using like `Attribute.FORECAST`.

As far as the class name, it wants the word "Utility" (or one of the other it listed) in the class name since it's under a path called "util" but that's not a naming convention this project uses so I wouldn't worry about that one.

zxue-usgs:
> Overall these changes look good! One thing I'd take a look at though is whether or not the click-to-identify is working as you would expect it to. For my categorical stuff to show up properly I had to add the new attribute I added to the list here:
> 
> https://github.com/USGS-CIDA/coastal-hazards/blob/master/coastal-hazards-portal/src/main/webapp/js/cch/objects/LayerIdentifyControl.js#L520

Thanks for the heads up, click-to-identify does not work. I will look into how we want it to be displayed. Holding off on this PR for now.

zxue-usgs:
I don't understand why travis failed.

zmoore-usgs:
Huh, yeah that's a weird one.. I'm thinking Travis might've gotten an error page trying to pull some dependencies (hence the `<html>` bit it mentioned not recognizing). Going to try to manually kick it off again

PR #1719, opened 2020-09-23
Gitlab merge

zmoore-usgs:
No need to actually review anything here since it has been reviewed in GitLab, just tagged you as an FYI @zxue-usgs 

Once your current PR is merged in I'll merge this in to get the GitHub repo here up-to-date with what's in GitLab and then I'll create my PR for migrating Rserve functionality to Java. That'll be the last PR we do here on GitHub and then we'll migrate over to working off of the GitLab repo.

I'd recommend that once we switch from this repo to the GitLab repo you just delete your local repo folder and re-clone from a fork of the GitLab one.

Some things to know about the GitLab repo:

It uses a different Git workflow. Instead of the forking workflow we use here (fork --> make changes and push to fork --> PR from fork to canonical master branch) it uses a branching workflow. You can still fork it and have your fork be your primary repo that you clone locally, however when pushing changes back up to canonical you won't make a PR from your fork into canonical. Instead, you'll want to push your changes up to a new branch on canonical itself. Also note that instead of developing off of "master" canonical uses tier-specific branches. So there's "development" for the current state of the code on the dev tier and "production" for the current state of the code on the prod tier. Ivan is currently working on getting QA setup so that should exist sometime soon as well. So for example (changed Gitlab URL for the example):

```
# Fork git@gitlab.host.gov:cmgp/coastal-hazards.git to personal account and clone locally
git clone git@gitlab.host.gov:zmoore-usgs/coastal-hazards.git

# Create branch for whatever thing I'm working on
git checkout -b new-feature

# Pull current development branch into feature branch
git pull canonical development

# Now my local feature branch is up-to-date with current dev code, so make changes....

# Changes made so stage, review, and commit
git add -A
git status
git commit -m "Made some cool changes"

# If something new is pushed into development while I'm still working I can just pull it again
git pull canonical development

# Push changes up to my fork periodically so they're backed up in case my laptop dies or something
git push origin new-feature:new-feature (pushes "new-feature" branch locally to "new-feature" branch on my fork at GitLab)

# Continue to make changes....

# Once changes are ready to be made into a pull-request push my feature branch up to canonical instead of my fork
git push canonical new-feature:zmoore-new-feature (pushes "new-feature" branch locally to "zmoore-new-feature" branch on the canonical repo at GitLab)

[...]

PR #1720, opened 2020-09-23
Port remaining Rserve and 52N functionality to Java

zmoore-usgs:
Once https://github.com/USGS-CIDA/coastal-hazards/pull/1719 is merged this should be whittled down to only the commits that are actually meaningful, so I wouldn't worry about looking at this until then.

- Port real time storms child item metadata parsing from 52N + Rserve to Java
- Port thumbnail generation from 52N + Rserve to Java
- Fixed a bug with Aggregation creation where if you created a new aggregation on the UI while looking at another aggregation it would copy the children to the new aggregation


zmoore-usgs:
Going to close and re-open this so it has the right set of commits

PR #1721, opened 2020-09-24
Port remaining Rserve and 52N functionality to Java

zmoore-usgs:
- Port real time storms child item metadata parsing from 52N + Rserve to Java
- Port thumbnail generation from 52N + Rserve to Java
- Fixed a bug with Aggregation creation where if you created a new aggregation on the UI while looking at another aggregation it would copy the children to the new aggregation


coastal-hazards-portal/src/main/webapp/WEB-INF/jsp/ui/info/index.jsp
```diff
@@ -155,7 +155,7 @@
 					</p>
 
 					<a href="https://marine.usgs.gov/coastalchangehazards/" target="_blank">USGS National Assessment of Coastal Change Hazards</a>
-					<a href="https://github.com/USGS-CIDA/coastal-hazards" target="_blank">Review the open source software behind the Portal available on Github</a>
+					<a href="https://www.usgs.gov/core-science-systems/national-geospatial-program/national-map" target="_blank">Thumbnail basemaps provided by the USGS National Map</a>
```
mhines-usgs, line 158: 🎉

mhines-usgs:
very impressive work, and thanks for writing up all the details about working with gitlab and the special branching methods. are there going to be any follow up tasks related to the infrastructure that will need to happen when this is merged (e.g. related to the pipelines/etc/ and dropping rserve/r/52n)?

zmoore-usgs:
Yeah, but it's more of a "can happen" vs. a "need to happen". Things will still work if Rserve and 52N are running in CHS, they just won't be used. But we would be able to completely remove them from the repo and pipelines once this is merged so a follow-up cleanup task definitely makes sense, just doesn't _have_ to be done necessarily

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/domain/DataDomainUtility.java
```diff
@@ -18,34 +19,34 @@
...
-		    Set<String> childDomain = retrieveDomainFromWFS(child); // recurse (again, avoiding cycles is important)
-		    domain.addAll(childDomain);
-		}
+                if(displayedIds.contains(child.getId())){
+                    Set<String> childDomain = retrieveDomainFromWFS(child); // recurse (again, avoiding cycles is important)
+                    domain.addAll(childDomain);
+                }
             }
-        } else if (item.getItemType() == Item.ItemType.data) {
+        } else if (item.getItemType() == Item.ItemType.data && Attributes.getDomainAttrs().contains(item.getAttr().toUpperCase())) {
```
isuftin, line 38: Does Attributes.getDomainAttrs() always return uppercased strings?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/domain/DataDomainUtility.java
```diff
@@ -18,34 +19,34 @@
...
             WFSService service = item.fetchWfsService();
             try {
                 Set<String> domainValuesAsStrings = client.getDomainValuesAsStrings(service, item.getAttr());
                 domain.addAll(domainValuesAsStrings);
             } catch (IOException e) {
-                log.error("unable to get domain from wfs", e);
+                log.error("Unable to get domain from wfs", e);
+            } catch(UnsupportedOperationException e) {
+                log.error("Unable to get domain for item :" + item.getId() + "(attr: " + item.getAttr() + ")", e);
```
isuftin, line 46: Consider using string formatting in error logging.

Is catching and not dealing with an UnsupportedOperationException really ok here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/gson/adapter/ItemTreeAdapter.java
```diff
@@ -24,7 +28,12 @@ public JsonElement serialize(Item src, Type typeOfSrc, JsonSerializationContext
 			JsonObject json = (JsonObject) fullItem;
 			simplified.add("id", json.get("id"));
 			simplified.add("itemType", json.get("itemType"));
-			simplified.add("title", json.getAsJsonObject("summary").getAsJsonObject("full").get("title"));
+			if(json.getAsJsonObject("summary") != null && json.getAsJsonObject("summary").getAsJsonObject("full") != null) {
```
isuftin, line 31: If `json.get("...")` returns null, are you sure that `json.getAsJsonObject("...")` also returns null? You can also do:

`if (json.has("summary") &&  json.getAsJsonObject("summary").has("full") { ... }` right?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
```diff
@@ -63,17 +74,31 @@ public void setBbox(double minX, double minY, double maxX, double maxY) {
...
+			Matcher matcher = pattern.matcher(source.getBbox());
 			if (matcher.matches()) {
 				double minX = Double.parseDouble(matcher.group(1));
 				double minY = Double.parseDouble(matcher.group(2));
 				double maxX = Double.parseDouble(matcher.group(3));
 				double maxY = Double.parseDouble(matcher.group(4));
-				envelope = new Envelope(minX, maxX, minY, maxY);
+				Pair<Double,Double> minPoint = new ImmutablePair<>(minX,minY);
+				Pair<Double,Double> maxPoint = new ImmutablePair<>(maxX,maxY);
+				return new ImmutablePair<>(minPoint, maxPoint);
```
isuftin, line 97: Note that you don't really use these variables except to return them so you could just:

```java
return new ImmutablePair<>(
  new ImmutablePair<Double,Double>(minX,minY),
  new ImmutablePair<Double,Double>(maxX,maxY)
);
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
```diff
@@ -99,4 +124,82 @@ public static Polygon envelopeToPolygon(Envelope e) {
...
+			calc.setDestinationGeographicPoint(points.getLeft().getLeft(), points.getRight().getRight());
+			double distLong = calc.getOrthodromicDistance();
+			double angleLong = calc.getAzimuth();
+
+			// Extend short dimension
+			if(distLong > distLat) {
+				// Extend latitude 50% in both directions
+				log.debug("Expand lat: " + distLat + " | " + distLong);
+				double expandDist = (distLong - distLat)/2;
+				calc = new GeodeticCalculator();
```
isuftin, line 153: calc can be initialized above the if block

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
```diff
@@ -99,4 +124,82 @@ public static Polygon envelopeToPolygon(Envelope e) {
...
+				calc.setStartingGeographicPoint(points.getLeft().getLeft(), points.getLeft().getRight());
+				calc.setDirection(angleLat, -expandDist);
+				double newMinX = calc.getDestinationGeographicPoint().getX();
+				calc.setStartingGeographicPoint(points.getRight().getLeft(), points.getLeft().getRight());
+				calc.setDirection(angleLat, expandDist);
+				double newMaxX = calc.getDestinationGeographicPoint().getX();
+				newBox.setBbox(newMinX, points.getLeft().getRight(), newMaxX, points.getRight().getRight());
+			} else if(distLat > distLong) {
+				// Extend longitude 50% in both directions
+				log.debug("Expand long: " + distLat + " | " + distLong);
```
isuftin, line 163: Should this log message reverse distLat and distLong? It's the same as in 151

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/MetadataResource.java
```diff
@@ -49,117 +25,19 @@
...
-			response = Response.ok(GsonUtil.getDefault().toJson(grouped, Map.class)).build();
-			
-		} catch (Exception e) {
-			log.error("Failed to parse metadata xml document. Error: " + e.getMessage() + ". Stack Trace: " + e.toString());
-		}
-
-		return response;
-	}
+		Response response;
+		ParsedMetadata result = MetadataUtil.parseMetadataXmlFile(postBody);
```
isuftin, line 33: Chances of exceptions here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/MetadataResource.java
```diff
@@ -49,117 +25,19 @@
...
-		if (!readFile.exists()) {
-			return Response.status(Response.Status.NOT_FOUND).build();
-		}
-		else if (!readFile.canRead()) {
-			responseContent.put("message", "Metadata Not Accessible");
-			return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity(responseContent).build();
-		}
-		else {
-			return Response.ok(IOUtils.toString(new FileInputStream(readFile)), MediaType.APPLICATION_XML_TYPE).build();
+		if(result != null) {
```
isuftin, line 35:
```java
if (result != null) {
  return Response.ok(GsonUtil.getDefault().toJson(result)).build();
}
return Response.status(HttpStatus.SC_INTERNAL_SERVER_ERROR).build();
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -196,13 +184,13 @@ protected String updateStormAlias(String alias, AliasManager aliasMan, String te
 		}
 	}
 	
-	protected Response instantiateStormAndHandleResponse(String templateId, String childJson, String alias, boolean active, AliasManager aliasMan, ItemManager itemMan){
+	protected Response instantiateStormAndHandleResponse(Document metadataXml, ParsedMetadata parsedMetadata, String templateId, String childJson, String alias, boolean active, AliasManager aliasMan, ItemManager itemMan){
 		Response response;
 		if (StringUtils.isEmpty(templateId)) {
 			//this is the server's fault rather than the client's
 			response = Response.status(500).build();
```
isuftin, line 191:
```java
return Response.status(500).build();
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -311,19 +305,19 @@ public Response instantiateStormTemplate(
 					if ("item".equalsIgnoreCase(copyType) || "alias".equalsIgnoreCase(copyType)) {
 						summary = copyExistingSummary(copyType, copyVal, itemMan, aliasMan);
 					} else {
-						summary = StormUtil.buildStormTemplateSummary(title, srcUsed);
+						summary = StormUtil.buildStormTemplateSummary(parsedMetadata.getTitle(), parsedMetadata.getSrcUsed());
 					}
 					if (null == summary) {
 						response = Response.status(HttpStatus.SC_BAD_REQUEST).build();
```
isuftin, line 311: Might want to log this

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -529,8 +519,10 @@ protected Summary gatherTemplateSummary(Summary previousSummary, List<Item> chil
 				Set<String> childKeywords = keywordsFromString(item.getSummary().getKeywords());
 				keywordSet.addAll(childKeywords);
 				List<Publication> childPubs = item.getSummary().getFull().getPublications();
-				for (Publication pub : childPubs) {
-					publicationSet.add(Publication.copyValues(pub, null));
+				if(childPubs != null && !childPubs.isEmpty()) {
```
isuftin, line 522: Do we have access to `StringUtils.isNullOrEpty()` here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -127,6 +127,13 @@ private static Service wmsService(String layer) {
 		return service;
 	}
 
+	public static String buildGeoServerWMSRequest(String layerParam, String bboxParam, String sldParam, String sizeParam, String crsParam) {
```
isuftin, line 130: This would read much, much easier if you did string format here

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/HttpUtil.java
```diff
@@ -67,4 +71,22 @@ public static String fetchDataFromUri(String endpoint) {
...
 		return data;
 	}
+
+	public static BufferedImage fetchImageFromUri(String endpoint) {
+
+		try(CloseableHttpClient client = HttpClientBuilder.create().build()) {
+			HttpUriRequest req = new HttpGet(endpoint);
+			log.debug("FetchImage - About to grab image from URL: " + endpoint);
+			HttpResponse response = client.execute(req);
+			try(ByteArrayInputStream bytes = new ByteArrayInputStream(EntityUtils.toByteArray(response.getEntity()))) {
```
isuftin, line 81: Weird nested try/catch. Shouldn't this just rethrow to line 86 if you don't include the nested catch on line 83? Or just throw e on line 84?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/MetadataUtil.java
```diff
@@ -60,16 +41,70 @@
...
-		props = JNDISingleton.getInstance();
-		cchn52Endpoint = props.getProperty("coastal-hazards.n52.endpoint");
-	}
+    public static Document parseMetadataBody(String postBody) {
+        try {
+            Document doc = null;
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            doc = factory.newDocumentBuilder().parse(new InputSource(new StringReader(postBody)));
+            doc.getDocumentElement().normalize();
+            return doc;
```
isuftin, line 53:
```java
return factory.newDocumentBuilder().parse(new InputSource(new StringReader(postBody))).
  getDocumentElement().normalize();
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/MetadataUtil.java
```diff
@@ -60,16 +41,70 @@
...
+            metadata.setData(data);
+            metadata.setPublications(publication);
+            metadata.setResources(resource);
+
+            // Only some, generally raster, metadata xml files will include EPSG data
+            try {
+                String epsgCode = CRS.lookupIdentifier(getCrsFromFgdcMetadata(postBody), true);
+                metadata.setEPSGCode(epsgCode);
+            } catch (Exception e) {
+                log.info("Unable to extract an EPSG code from metadata XML; This is not an error. Returning null. Reason: " + e.getMessage());
```
isuftin, line 98: You don't return null here. You're returning an object on line 101 which has already been initialized

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ThumbnailUtil {
+	private static final Logger log = LoggerFactory.getLogger(GeoserverUtil.class);
+	private static final DynamicReadOnlyProperties props;
+	private static final String portalPublicUrl;
+    private static final String thumbanilBaseWmsUrl;
```
isuftin, line 30: weird formatting

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+            if(baseMap == null) {
+                baseMap = new BufferedImage(THUMBNAIL_SIZE, THUMBNAIL_SIZE, BufferedImage.TYPE_INT_RGB);
+                baseMap.getGraphics().fillRect(0, 0, THUMBNAIL_SIZE, THUMBNAIL_SIZE);
+            }
+            
+            // Draw generated thumb onto basemap
+            if(itemThumb != null) {
+                baseMap.getGraphics().drawImage(itemThumb, 0, 0, null);
+            } else {
+                log.error("Item " + item.getId() + " thumbnail failed to generate. Returning blank basemap.");
```
isuftin, line 79: Did you mean to return something here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+                    }
+                } else {
+                    log.warn("Supplied template or aggregation: " + item.getId() + " has no visible child items for thumbnail generation.");
+                }
+            } else {
+                log.debug("Reched max child depth when rednering thumbnail for: " + selectedItem);
+            }
+        } else {
+            Service proxyWms = null;
+            if(item.getServices() != null && !item.getServices().isEmpty()) {
```
isuftin, line 142: You only really care about the null check here

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/domain/DataDomainUtility.java
zmoore-usgs: Yeah, but by convention not standardization. I could add an explicit conversion here to be safe

isuftin: Might be a good idea

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/domain/DataDomainUtility.java
zmoore-usgs: I believe so. The result is returning an empty domain instead of throwing an exception. This was an issue when trying to build an SLD for an Aggregation that contained a Shoreline item and a non-Shoreline item. The SLD would return a 500 because this would try to calculate domain values for items that didn't support it (which would throw the UnsupportedOperationException). Now instead of that error causing the full SLD generation to fail it just results in no Domain being generated for the unsupported items instead of an exception being thrown.

isuftin: Sounds good. Thanks

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/gson/adapter/ItemTreeAdapter.java
zmoore-usgs: I was getting an NPE here before and this fixed it so I assume that this works as expected, but I think your option is cleaner I wasn't aware of `json.has()`

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
zmoore-usgs: Yeah, just wasn't sure if nested constructors like that would be too gross looking but I think that looks fine

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
zmoore-usgs: Whoops, yep

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/gson/adapter/ItemTreeAdapter.java
isuftin: To be fair, I don't know whether the gson lib returns true if there is a null hanging out in there. Something to try

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/MetadataResource.java
zmoore-usgs: Will double check, but this is just running the exact same block of code just extracted to a util function so shouldn't really be any different than it was before

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
zmoore-usgs: Yeah I pretty much 1-to-1 copied stuff from R for these, but string builders or string formatting would be cleaner. I can take a look at that

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/HttpUtil.java
zmoore-usgs: I was wanting to capture both autoclosables but I could just use a try instead of try/catch for this internal one I think

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
zmoore-usgs: No, at this point the only thing that has been drawn onto the image is the basemap so that's effectively what happens. The log is just to inform that instead of returning a blank image or an error for a failed thumbnail it'll return what has already been drawn on the image up to this point which is the basemap

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/MetadataUtil.java
zmoore-usgs: Yeah this is the same existing code that was just moved from the MetadataResource into here but you're right, it doesn't. Will tweak that log message to match reality

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/HttpUtil.java
isuftin: I'm pretty sure that it would have the same effect

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/MetadataResource.java
zmoore-usgs: The parseMetadataXmlFile method should handle all exceptions and return null

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/MetadataUtil.java
zmoore-usgs: .normalize() has a void return type funfortunately

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
```diff
@@ -99,4 +122,84 @@ public static Polygon envelopeToPolygon(Envelope e) {
...
+				DirectPosition minPoint = new DirectPosition2D(sourceCRS, points.getLeft().getLeft(), points.getLeft().getRight());
+				minPoint = transform.transform(minPoint, null);
+				DirectPosition maxPoint = new DirectPosition2D(sourceCRS, points.getRight().getLeft(), points.getRight().getRight());
+				maxPoint = transform.transform(maxPoint, null);
+				newBox.setBbox(minPoint.getCoordinate()[0], minPoint.getCoordinate()[1], maxPoint.getCoordinate()[0], maxPoint.getCoordinate()[1]);
+			} catch(Exception e) {
+				log.error("Failed to transform BBOX points from EPSG:4326 to " + targetCRS);
+				newBox = new Bbox();
+			}
+			log.debug("Transformed box: " + newBox.getBbox());
```
isuftin, line 200: If I follow this logic, if you hit an exception and fall into line 200, won't you get a an NPE here?

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -196,36 +184,37 @@ protected String updateStormAlias(String alias, AliasManager aliasMan, String te
...
+
+		Response response = instantiateTemplate(metadataXml, parsedMetadata, templateId, childJson);
+
+		if (response.getStatus() == HttpStatus.SC_OK) {
+			if (active) {
+				hoistNewTemplateToTopLevel(templateId, itemMan);
+			}
 
-			if (response.getStatus() == HttpStatus.SC_OK) {
+			if (!StringUtils.isEmpty(alias)) {
```
isuftin, line 200: `if (StringUtils.isNotEmpty(alias)) {...}`

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
```diff
@@ -196,36 +184,37 @@ protected String updateStormAlias(String alias, AliasManager aliasMan, String te
...
+				String originalTemplateId = updateStormAlias(alias, aliasMan, templateId);
```
isuftin, line 201: This variable isn't used outside of the if scope below so that can be moved into it

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
isuftin, line 200: This can be changed to be:

```java
if (StringUtils.isNotEmpty(alias) && active) {
  orphanOriginalStormTemplate(
    updateStormAlias(alias, aliasMan, templateId), 
    itemMan
 );
}
```

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/model/Bbox.java
zmoore-usgs: Shouldn't because of the copy on line 184. Should end up logging whatever the original bbox was

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/gson/adapter/ItemTreeAdapter.java
zmoore-usgs: That seems to work as well

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ThumbnailResource.java
```diff
@@ -71,6 +75,31 @@ public Response getImage(@PathParam("id") String id, @Context Request request) {
 		return response;
 	}
 
+	@GET
+	@Path("/item/{id}/generate")
+	@Consumes(MediaType.TEXT_PLAIN)
+	@Produces(MediaType.APPLICATION_JSON)
+	public Response genImage(@PathParam("id") String id, @Context HttpServletRequest request) {
+		Response response = null;
```
isuftin, line 83: No need to init this to null since you will assign to it below.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/ThumbnailResource.java
```diff
@@ -71,6 +75,31 @@ public Response getImage(@PathParam("id") String id, @Context Request request) {
...
+		try(ItemManager manager = new ItemManager()) {
+			Item item = manager.load(id);
+
+			if(item == null) {
+				response = Response.status(HttpStatus.SC_NOT_FOUND).build();
+			} else {
+				String imageBase64 = ThumbnailUtil.generateBase64Thumbnail(item);
+
+				if(imageBase64 != null && !imageBase64.isEmpty()) {
```
isuftin, line 92: `if (imageBase64 != null && imageBase64.isNotEmpty()) ...`

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/GeoserverUtil.java
```diff
@@ -127,6 +127,15 @@ private static Service wmsService(String layer) {
...
+		String url = geoserverInternalEndpoint + (geoserverInternalEndpoint.endsWith("/") ? "" : "/") + "proxied/wms?";
+		String params = String.format(
+			"FORMAT=image/png&TRANSPARENT=TRUE&STYLE=cch&VERSION=1.3.0&SERVICE=WMS&EXCEPTIONS=application%2Fvnd.ogc.se_blank&REQUEST=GetMap&%s&%s&%s&%s&%s",
```
isuftin, line 133: Be careful here because I do have the IDE saying that `F` is an unknown format specifier. Assuming it's the percent sign in here somewhere. Just verify it works for you before shipping.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ThumbnailUtil {
+	private static final Logger log = LoggerFactory.getLogger(GeoserverUtil.class);
+	private static final DynamicReadOnlyProperties props;
+	private static final String portalPublicUrl;
```
isuftin, line 29: Consider switching this a constants format of caps + underscore

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+	private static final String thumbanilBaseWmsUrl;
+	private static final String thumbnailCRS;
+	private static final Integer MAX_THUMBNAIL_CHILD_DEPTH = 4;
+	private static final Integer THUMBNAIL_SIZE = 150;
+	
+	static {
+		props = JNDISingleton.getInstance();
+		portalPublicUrl = props.getProperty("coastal-hazards.public.url");
+		thumbanilBaseWmsUrl = props.getProperty("coastal-hazards.thumbnail.basemap.wms");
+		thumbnailCRS = "EPSG:3857";
```
isuftin, line 39: Wondering why not just set this on line 31?

zmoore-usgs: Yeah, was originally going to make this configurable but then decided it didn't need to be. Will change that

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+			log.error("Failed to generate BBox for item: " + item.getId(), e);
+		}
+
+		return null;
+	}
+
+	private static Pair<BufferedImage,Integer> generateThumbnail(Item item, String selectedItem, String bboxVal, Integer ribbonNum, ItemManager manager, Integer depth) {
+		Pair<BufferedImage,Integer> result = new ImmutablePair<>(null,ribbonNum);
+
+		if(item.getItemType().equals(ItemType.uber)) {
```
isuftin, line 99: All of this might read easier as a switch statement.

```java

            switch (item.getItemType()) {
                case uber:
                    log.error("Thumbnail generation is not supported for uber.");
                    break;
                case template:
                case aggregation:
                    log.debug("Supplied item is of type: " + item.getItemType() + ". Generating thumbnail from visible children..");
                    // Exit case
                    if(depth < MAX_THUMBNAIL_CHILD_DEPTH) {
                        List<String> itemIds = item.getDisplayedChildren();
                        
                        if(item.isShowChildren() && !itemIds.isEmpty()) {
                            List<BufferedImage> childThumbnails = new ArrayList<>();
                            for(String itemId : itemIds) {
                                Item childItem = manager.load(itemId);
                                if(childItem != null) {
                                    Pair<BufferedImage,Integer> childResult = generateThumbnail(childItem, selectedItem, bboxVal, ribbonNum, manager, depth + 1);
                                    
                                    if(childResult.getLeft() != null) {
                                        childThumbnails.add(childResult.getLeft());
                                    }
                                    
                                    ribbonNum = childResult.getRight();
                                }
                            }
                            
                            // Merge child thumbnails
                            if(!childThumbnails.isEmpty()) {
                                BufferedImage aggThumb = new BufferedImage(THUMBNAIL_SIZE, THUMBNAIL_SIZE, BufferedImage.TYPE_INT_ARGB);
                                Graphics g = aggThumb.getGraphics();
                                
                                for(BufferedImage childThumb : childThumbnails) {
                                    g.drawImage(childThumb, 0, 0, null);
                                }
                                
                                result = new ImmutablePair<>(aggThumb, ribbonNum);
                            }
                        } else {
                            log.warn("Supplied template or aggregation: " + item.getId() + " has no visible child items for thumbnail generation.");
                        }
                    } else {
                        log.debug("Reched max child depth when rednering thumbnail for: " + selectedItem);
                    }
                    break;
                default:
                    Service proxyWms = null;
                    if(item.getServices() != null) {
                        for(Service service : item.getServices()) {
                            if(service.getType().equals(Service.ServiceType.proxy_wms)) {
                                proxyWms = service;
                                break;
                            }
                        }
                    }
                    if(proxyWms != null) {
                        String crsParam = "CRS=" + thumbnailCRS;
                        String layerParam = "LAYER=" + proxyWms.getServiceParameter();
                        String bboxParam = "BBOX=" + bboxToWMSParam(bboxVal);
                        String sldParam = "SLD=" + portalPublicUrl + "/data/sld/" + item.getId();
                        sldParam += "?selectedItem=" + selectedItem;
                        sldParam += item.isRibbonable() ? "%26ribbon=" + ribbonNum : "";
                        String sizeParam = "WIDTH=" + THUMBNAIL_SIZE + "&" + "HEIGHT=" + THUMBNAIL_SIZE;
                        result = new ImmutablePair<>(
                                HttpUtil.fetchImageFromUri(GeoserverUtil.buildGeoServerWMSRequest(layerParam, bboxParam, sldParam, sizeParam, crsParam)),
                                ribbonNum + 1
                        );
                    } else {
                        log.error("Could not generate thumbnail for item: " + item.getId() + ". Item has no proxy WMS service.");
                    }
                    break;
            }
```

(thanks NetBeans)

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+					for(String itemId : itemIds) {
+						Item childItem = manager.load(itemId);
+						if(childItem != null) {
+							Pair<BufferedImage,Integer> childResult = generateThumbnail(childItem, selectedItem, bboxVal, ribbonNum, manager, depth + 1);
+
+							if(childResult.getLeft() != null) {
+								childThumbnails.add(childResult.getLeft());
+							}
+
+							ribbonNum = childResult.getRight();
```
isuftin, line 119: You're assigning to a method param here. Careful.

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+			if(item.getServices() != null) {
+				for(Service service : item.getServices()) {
+					if(service.getType().equals(Service.ServiceType.proxy_wms)) {
+						proxyWms = service;
+						break;
+					}
+				}
+			}
+
+			if(proxyWms != null) {
```
isuftin, line 151: Consider using string formatting here or at least doing a string concat() on these

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/util/ThumbnailUtil.java
```diff
@@ -0,0 +1,183 @@
...
+				);
+			} else {
+				log.error("Could not generate thumbnail for item: " + item.getId() + ". Item has no proxy WMS service.");
+			}
+		}
+
+		return result;
+	}
+
+	private static BufferedImage fetchBaseMap(String bboxVal) {
```
isuftin, line 171: String format or concat here

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/domain/DataDomainUtility.java
```diff
@@ -18,34 +19,32 @@
...
  */
 public class DataDomainUtility {
-    
     private static final Logger log = Logger.getLogger(DataDomainUtility.class);
 
     public static SortedSet<String> retrieveDomainFromWFS(Item item)  {
         SortedSet<String> domain = new TreeSet<>();
-        
+
         WFSGetDomain client = new WFSGetDomain();
```
isuftin, line 27: Below you may want to do:

```java
if (item == null) { throw new IllegalArgumentException("Item must be valid data item"); }

ItemType = item.getItemType()

...
```

And then do the rest of your logic because you do use `item.getItemType()` a number of times below

coastal-hazards-portal/src/main/java/gov/usgs/cida/coastalhazards/rest/data/TemplateResource.java
zmoore-usgs: Looking at this a bit more, that's not quite going to work. `updateStormAlias` makes modifications to the DB and doesn't just return the original item ID that matched the alias. We want to always update the alias if it's providing a new one, but we only want to take the item ID that the alias used to point to and and orphan it if the new storm is active. So I think the current logical flow makes sense

isuftin: Shouldn't Java first call updateStormAlias() before calling the outer orphanOriginalStormTemplate() function, making the two ways of writing this the same?

